PR #5, opened 2022-11-29
Cleanup and pull projections into numpy arrays

mcb64:
This is a first step to getting projection fits into the camera viewer.  The main work here is that the projection sums are now numpy arrays passed into pycaqtimage.pyCreateImageBuffer.  The C++ sip code checks that these are the right size and type, and if so, uses them to contain the sums.  Otherwise, internal buffers are allocated.

Step two would be to use matplotlib or the like to construct the projection bitmaps.  Then, fitted curves could easily be plotted onto these.  (Currently, an image is constructed in C++ and passed up.  Yuck.)


mcb64:
You had your chance while this was easy to review... :-)

klauer:
Click this thing and select at least one of ecs-pd members to ask for their review:
image

That is to say you usually have to request a review from someone before giving up without a reviewer...

PR #7, opened 2023-02-09
Black formatting, as requested by Ken Lauer.

mcb64:
No code was changed in the making of this pull request. :-)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,29 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.3.0
+    hooks:
+    -   id: trailing-whitespace
+    -   id: end-of-file-fixer
+    -   id: check-ast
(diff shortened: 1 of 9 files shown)


klauer (approved):
The requester approves! This makes things a lot more readable and consistent. Thanks.

It might also be a good idea to add in a pre-commit configuration like we use in all other projects. It'll reformat-on-commit when required.

This linked one does `black` and `isort` (import sorting/cleaning), along with other linter checks that would assuredly fail in this project (so maybe we skip those for now):
https://github.com/pcdshub/ads-async/blob/d68246c7a8d59ed8a0cee8c2dfba8815c43bc5a4/.pre-commit-config.yaml#L25-L36

It's pretty simple, with just 
```
$ pre-commit install
$ git commit  # commit as usual, fixes applied as needed
```

mcb64:
isort was unhappy installing, so I removed it.  And I gather I want to commit to *my* branch, so that hook gets removed...

klauer:
The "don't commit to master" thing is saying you should make a feature branch for what you're working on. Like, "style_black" or "refactor_pyqt" or something.

No big deal if you want to continue with `master` on your end. 
`isort` can be pretty nice - but also your choice 👍

PR #8, opened 2023-02-09
Add lineouts and (super)gaussian curve fits.

mcb64:


klauer:
Is this ready for review?

mcb64:
Right, didn't assign anyone, did I?

ProjWidget.py
```diff
@@ -154,7 +328,69 @@ def makeImage(self, xminR, xmaxR, yminR, ymaxR):
...
+                self.gui.ui.display_image.lMarker[2].oriented(),
+                self.gui.ui.display_image.lPenColor[2],
+            )
+        if self.gui.ui.checkBoxM4Lineout.isChecked():
+            (ymin, ymax) = self.plotLineout(
+                ax,
+                self.is_x,
+                linelim,
+                x,
+                xmin,
```
klauer: Is `idx` a missing parameter here by chance? 

My eyes were drawn to the 10ish positional parameters and this seemed to not fit the pattern of the above, but I can't really tell. Might consider a loop to get rid of the copy/paste, if so

camviewer_ui_impl.py
```diff
@@ -2712,4 +2768,25 @@ def getConfig(self):
...
+            l = [ll == "1" for ll in self.cfg.projfittype]
+            if l[0]:
+                self.ui.radioGaussian.setChecked(True)
+            if l[1]:
+                self.ui.radioSG4.setChecked(True)
+            if l[2]:
+                self.ui.radioSG6.setChecked(True)
+            self.calib = float(self.cfg.projcalib)
+            self.ui.lineEditCalib.setText(str(self.calib))
+        except:
```
klauer: `try:` with a really big block and a bare `except: pass` can make debugging weird issues pretty frustrating...

`except:` is generally recommended to be `except Exception:` instead, because the former will eat up things like `KeyboardInterrupt` but the latter will not. I'd imagine this try block isn't intending to catch that - see [exception hierarchy](https://docs.python.org/3/library/exceptions.html#exception-hierarchy))

Consider maybe at least dumping out a traceback with a `logger`:

```python
try:
    do_the_thing
except Exception:
    logger.exception("Failed to do_the_thing")
```

That assumes a logger exists at module-level like `import logging; logger = logging.getLogger(__name__)` or some other way. Alternatively, an `import traceback` and in the except block `traceback.print_exc()` could be better than nothing.

camviewer_ui_impl.py
```diff
@@ -2470,13 +2499,44 @@ def dumpConfig(self):
                     % (i + 1, lMarker[i].abs().x(), lMarker[i].abs().y())
                 )
             g.write("dispspec    " + str(self.dispspec) + "\n")
+            f.write(
```
klauer, line 2511: This config thing is some weird format made up for camviewer, right? This thing reads it?
https://github.com/mcb64/camviewer/blob/0bf97cb79cf1476e06b935b7274b5a129b1b750f/camviewer_ui_impl.py#L78-L103

What's the difference between an oldcfg and a newcfg?
Am I understanding it correctly that it'll load the old format, write the new format, and then delete the old one?

ProjWidget.py
```diff
@@ -5,6 +5,111 @@
...
+
+def sg4(x, amplitude=1.0, center=0.0, width=1.0):
+    return amplitude * np.exp(-2.0 * ((x - center) ** 4 / max(tiny, width**4)))
+
+
+def sg6(x, amplitude=1.0, center=0.0, width=1.0):
+    return amplitude * np.exp(-2.0 * ((x - center) ** 6 / max(tiny, width**6)))
+
+
+class SG4Model(Model):
```
klauer, line 25: Where'd these come from? Copied/modified from lmfit's models?

ProjWidget.py
mcb64: Absolutely correct.  I've modified this in another commit to use a loop.  Meanwhile, found another bug while testing this... my code for deciding whether or not to display a lineout was looking at the wrong dimension.

It's always something...

camviewer_ui_impl.py
mcb64: At this point, there is no logger.

And this block is in my state restore code anyway... so if it fails, I don't really care.  That's why I didn't really make any effort to try to figure out exactly why it failed and debug it... just move on and start displaying the image.

camviewer_ui_impl.py
mcb64: Yes.  I keep state files for each camera you view in ~/.yagviewer/BLAH.  So it will remember your orientation, your marker settings, etc. etc.

Originally back when dinosaurs roamed the earth, this was just the PV base name.  But as camera types proliferated, I decided that this should be PVNAME,TYPE, so if a camera gets upgraded or otherwise changed, the old state is ignored.  (This actually happened and caused camviewer crashes.)  Now, that we've decided to drop everything except area detector, I'm back to just PV base name.

However, I think I have told you that I am an evangelist for the First Church of Backwards Compatibility... so it would be heresy to just ignore the existing PVNAME,GE files!  So this code tries to find an old one, read it in... and delete it, so the user sins no more...

It's the same format, just a different name.

ProjWidget.py
mcb64: Yes.  lmfit had a gaussian model that evidently played games to avoid problems with small numbers.  I thought these guys probably knew what they were doing, so I basically copied their gaussian model as a template and changed the computation to an sg4 or sg6 model.

camviewer_ui_impl.py
mcb64: I'll change all of the bare excepts to except Exception though.

klauer:
So glad to see flake8 re-added 👍 

This flake8 config might be of interest - we use it in other projects to make things a bit more lenient:
https://github.com/pcdshub/cookiecutter-pcds-python/blob/master/%7B%7B%20cookiecutter.folder_name%20%7D%7D/.flake8

mcb64:
Yeah, I had figured out the E501, which is why I took out flake to begin with.  Might as well make things standard though...

ZLLentz:
Somewhat unrelated to this code itself: is there any easy way to feed generated images to camviewer to check out the curve fitting? We're a bit lacking in real beam to try gaussian fits on.

mcb64:
Not any more.  I used to be able to use random XTC files, but I removed this.

When I tested, I typically just used a bright spot on a random cam image and zoomed in the ROI.  I trust lmfit is doing the right fit... the only question is if I am displaying it properly, and it certainly looks like something is being fitted.

mcb64:
Just FYI, the jira ticket (https://jira.slac.stanford.edu/browse/LCLSECSD-1752) has a lot of fitted screenshots using this approach.

ZLLentz:
I agree, in those images it 100% looks like it's doing the right thing

dialogs.py
```diff
@@ -237,18 +246,18 @@ def __init__(self, dir, gui, parent=None):
 
                 pidlabel = QLabel(self)
                 pidlabel.setTextFormat(QtCore.Qt.RichText)
-                pidlabel.setText(pre + l[1] + post)
+                pidlabel.setText(pre + ll[1] + post)
                 self.gridLayout.addWidget(pidlabel, i, 1, 1, 1)
 
-                l = open(dir + file).readlines()
+                lines = open(dir + file).readlines()
```
ZLLentz, line 252: This isn't really the functional change part of this PR but I notice this file never gets closed

ZLLentz (approved):
I'm not sure what else to add, it seems to work and Ken already found most of the things

PR #9, opened 2023-02-16
Fix projections for color images

mcb64:
If the image is color, we still want to work on the grayscale image for projections, otherwise we get into math error problems.

Also, close a file, as Zach pointed out.

### DIFF
--- a/pycaqtimage/pycaqtimage.sip
+++ b/pycaqtimage/pycaqtimage.sip
@@ -546,7 +546,7 @@ static void _computeRoiProj(ImageBuffer* imageBuffer, QRectF* rectRoi, bool bPro
 	uint32_t* pPixel = pPixelLineStart;
 	for (int iX = x1; iX <= x2; ++iX, ++pPixel) {
 #define SUMRGB(x) (((x)&0xff)+(((x)>>8)&0xff)+(((x)>>16)&0xff))
-	    int iValue = isColor ? SUMRGB(*pPixel) : *pPixel;
+	    uint32_t iValue = isColor ? SUMRGB(*pPixel) : *pPixel;
 	    if ( iValue >= 0x10000) {
 		fprintf(stderr, "Pixel value (%d,%d) too large: value 0x%x\n", iX, iY, iValue);
 		continue;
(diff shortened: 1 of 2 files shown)


pycaqtimage/pycaqtimage.sip
```diff
@@ -540,11 +540,13 @@ static void _computeRoiProj(ImageBuffer* imageBuffer, QRectF* rectRoi, bool bPro
...
     uint64_t  u64PixelSqSum         = 0;
     uint32_t* pPixelLineStart       = pImgValue + y1 * width + x1;
+    int isColor = imageBuffer->isColor && !imageBuffer->useGray;
 
     for (int iY = y1; iY <= y2; ++iY, pPixelLineStart += width) {
 	uint32_t* pPixel = pPixelLineStart;
 	for (int iX = x1; iX <= x2; ++iX, ++pPixel) {
-	    int iValue = *pPixel;
+#define SUMRGB(x) (((x)&0xff)+(((x)>>8)&0xff)+(((x)>>16)&0xff))
+	    int iValue = isColor ? SUMRGB(*pPixel) : *pPixel;
```
klauer: Not unsigned/`uint32_t`?

pycaqtimage/pycaqtimage.sip
```diff
@@ -540,11 +540,13 @@ static void _computeRoiProj(ImageBuffer* imageBuffer, QRectF* rectRoi, bool bPro
     uint64_t  u64PixelSum           = 0;
     uint64_t  u64PixelSqSum         = 0;
     uint32_t* pPixelLineStart       = pImgValue + y1 * width + x1;
+    int isColor = imageBuffer->isColor && !imageBuffer->useGray;
 
     for (int iY = y1; iY <= y2; ++iY, pPixelLineStart += width) {
 	uint32_t* pPixel = pPixelLineStart;
 	for (int iX = x1; iX <= x2; ++iX, ++pPixel) {
-	    int iValue = *pPixel;
+#define SUMRGB(x) (((x)&0xff)+(((x)>>8)&0xff)+(((x)>>16)&0xff))
```
klauer, line 548: This says "sum the R G and B bytes of the pixel (the lower 24-bits of this unsigned integer) into a single value"

klauer (approved):
A small PR to camviewer? Inconceivable

Looks good to me 👍

pycaqtimage/pycaqtimage.sip
mcb64: Yep.  If you take a look a bit higher... 

[https://github.com/mcb64/camviewer/blob/c054558ad04f4cacf081ce63f43bf6410dc844ef/pycaqtimage/pycaqtimage.sip#L338-L339](url)

So are basically choosing to take something that we earlier did RGB and now do GRAY.

pycaqtimage/pycaqtimage.sip
mcb64: Point taken.

PR #10, opened 2023-03-08
Baseline fitting, Optionally save spatial calibration in a PV, empty GLOBAL fix.

mcb64:
Three changes:

- Don't crash if the GLOBAL configuration file is empty.
- Give an option to fit a baseline to the gausian/supergaussian fits.
- Add an Expert Mode option to connect the spatial calibration value to a PV.


klauer (approved):
Code looks fine. On the assumption that you've tested this and it works as you hoped it would, I'd say this is good to be merged 👍

mcb64:
Yes, I've tested it and it seems to work just fine.  Thanks!

PR #11, opened 2023-07-11
Fix up lensPV handling.

mcb64:
Very few cameras have lenses, so we had some *old*, non-working PV-handling code that tried to connect to the lens PV.  This fixes up the connection.


ZLLentz (approved):
My read of this is: the previous code simply doesn't work and nobody noticed for a while.

mcb64:
Your read is correct, sir.

PR #13, opened 2023-10-05
Adding Display Format feature

ctslac:
Adding a feature to allow the user to adjust the display format of Fid Widths FWHMH field.

This was tested by inputting several different combinations of test strings to ensure that the values were displayed correctly and checking the yagviewer/PVNAME file to make sure the setting was properly updating.

### DIFF
--- a/advanced.ui
+++ b/advanced.ui
@@ -93,7 +93,7 @@
    <item row="4" column="0">
     <widget class="QLabel" name="label_5">
      <property name="text">
-      <string>Display Format</string>
+      <string>FWHM Display Format</string>
      </property>
     </widget>
    </item>
--- a/advanced.ui
+++ b/advanced.ui
@@ -7,54 +7,17 @@
     <x>0</x>
     <y>0</y>
     <width>380</width>
-    <height>171</height>
+    <height>220</height>
    </rect>
   </property>
   <property name="windowTitle">
    <string>Dialog</string>
   </property>
   <layout class="QGridLayout" name="gridLayout">
-   <item row="0" column="0" colspan="2">
-    <widget class="QLabel" name="label">
-     <property name="text">
-      <string>Viewing Area Width (&gt; 450 pixels)</string>
-     </property>
-    </widget>
-   </item>
-   <item row="5" column="0">
-    <widget class="QPushButton" name="showevr">
-     <property name="minimumSize">
-      <size>
-       <width>0</width>
-       <height>33</height>
-      </size>
-     </property>
-     <property name="text">
-      <string>Open Evr</string>
-     </property>
-    </widget>
-   </item>
-   <item row="1" column="0" colspan="2">
-    <widget class="QLabel" name="label_3">
-     <property name="text">
-      <string>Viewing Area Height (&gt; 450 pixels)</string>
-     </property>
-    </widget>
-   </item>
    <item row="1" column="2">
     <widget class="QLineEdit" name="viewHeight"/>
    </item>
-   <item row="4" column="0" colspan="3">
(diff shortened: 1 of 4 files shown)


camviewer_ui_impl.py
```diff
@@ -2587,11 +2596,13 @@ def getConfig(self):
         self.cfg.add("projection", "0")
         self.cfg.add("markers", "0")
         self.cfg.add("dispspec", "0")
+        # self.cfg.add("displayFormat", "%12.8g")
```
mcb64: Remove this.  displayFormat isn't a global option, so you shouldn't put it here.

camviewer_ui_impl.py
```diff
@@ -2587,11 +2596,13 @@ def getConfig(self):
...
         if not self.cfg.read(self.cfgdir + "GLOBAL"):
             self.cfg.add("config", "1")
             self.cfg.add("projection", "1")
             self.cfg.add("markers", "1")
             self.cfg.add("dispspec", "0")
+            # self.cfg.add("displayFormat","%12.8g")
```
mcb64: Or here.

camviewer_ui_impl.py
```diff
@@ -2602,6 +2613,8 @@ def getConfig(self):
                 self.cfg.add("markers", self.options.marker)
             if self.options.camcfg is not None:
                 self.cfg.add("dispspec", self.options.camcfg)
+            # if self.options.displayFormat is not None:
```
mcb64: Now, self.options is the command line parameters that we have previously parsed.  It's not unreasonable to say you could pass in the displayFormat on the command line... but if you are going to do this, you need to change the option parser setup to include "displayFormat" as an acceptable option that takes a string argument.  Either do that, or remove this.  (I'd remove this... I think displayFormat is sufficiently obscure that it doesn't need to be a command line argument, but the choice is yours, and I'd approve either way.)

advanced.ui
```diff
@@ -74,16 +58,42 @@
...
+    <widget class="QLineEdit" name="projSize"/>
+   </item>
+   <item row="4" column="2">
+    <widget class="QLineEdit" name="displayFormat"/>
+   </item>
+   <item row="4" column="0">
+    <widget class="QLabel" name="label_5">
      <property name="text">
-      <string>Spatial Calibration PV name</string>
+      <string>Display Format</string>
```
mcb64: Maybe "Fit Display Format" or "FWHM Display Format"?  (Yeah, it's used for 1/e^2 width too, but "Display Format" is a little too generic, I think.)

mcb64 (dismissed):
I'd like to see the name in the advanced.ui change to something less generic than "Display Format", and comments should be removed before merging.

Should you decide that you would like displayFormat to be a command line parameter, I would like to see that code before you merge.

But if you just remove the comments and fix up the name, this should be good to go.

camviewer_ui_impl.py
ctslac: removed

camviewer_ui_impl.py
ctslac: removed

camviewer_ui_impl.py
ctslac: I think i will opt to just remove this

advanced.ui
ctslac: I went with the FWHM Display Format, I think its easier to understand.

mcb64 (dismissed):
Looks good to me.

mcb64 (approved):
Everything still looks good...

ZLLentz:
I'm going to edit the repo to not expect the standard status checks and then it will us merge

PR #14, opened 2024-04-10
Put orientation on main screen.

mcb64:
In addition to the orientation drop down, add a comboBox to the main screen to track the orientation.

### DIFF
--- a/camviewer.ui
+++ b/camviewer.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>1119</width>
-    <height>992</height>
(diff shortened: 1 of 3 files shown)


ZLLentz (approved):
Code looks normal and works as advertised ✔️ 

There's a small delta-t where the view of the camera I'm checking on goes to black until the next frame, but I suspect this was always a feature and not related to the new orientation selection.

mcb64:
Yeah, that's just the way it works, unfortunately.  Something for a future bug fix.  Or not... ;-)

PR #16, opened 2024-11-22
ENH: timeout rework

ZLLentz:
Previous behavior:
- The camviewer has 1 and 9 hour timeouts.
- After 1 hour, there is a pop-up that asks if you want to extend the timeout by 1 or 9 hours. Presumably this also appears after a 9 hour timeout.
- If a timeout expires, the gui stops getting images from the camera (but the camera continues to run), leaving a stale image.
- To re-establish connection, you can either swap the active cam back and forth or reconnect from the menu.

New behavior:
- Pop-up has been removed
- To reset the timeout: interact with the display in any way (click, type, etc.)
- Live-updating timeout display has been added
- The timer's starting value depends on the max display rate
- Max display rate can be changed live by the user
- 1 Hz = no timeout, ever
- 2-5 Hz = 7 days timeout
- 6-29 Hz = linear interp between 7 and 1 day
- 30 Hz+ = 1 day timeout
- If a timeout expires, rate drops to 1Hz until interacting with the GUI again (therefore: never a stale image if the real cam is running)
- pre-existing idle cli arg is now defunct (not sure what to do with it)
- min_timeout and max_timeout cli args added to help test the timeouts (timeouts in seconds, defaults to 1 day for the min and 7 days for the max)
- The existing "reconnect" menu action, whose original implementation is no longer valid, now disconnects from and reconnects to the IOC of the currently selected camera.

I've been testing this via passing `--min_timeout 10`, then setting the rate to 30Hz, then watching the timeout apply after 10 seconds.


KaushikMalapati:
Would it be unhelpful for me to leave comments before you make this ready for review?

ZLLentz:
> Would it be unhelpful for me to leave comments before you make this ready for review?

The main reason I didn't ask for reviews on this is because I'm waiting for feedback from the scientific staff. I didn't want to spend time cleaning up and involving other contributors until I was sure the laser scientists were happy with the feature set. See https://jira.slac.stanford.edu/browse/ECS-5976.

If you have comments you want to leave, feel free! All feedback is helpful! On this PR I'm especially interested in design feedback. But if the scientists come back and say "something different please" my resolution to a comment could end up being some deleted lines without a real resolution.

ZLLentz:
Ok, I've had enough positive feedback from stakeholders and I did a first-pass cleanup. I'm reasonably happy with this except for the half-baked cli args, but I'm also not motivated to clean up the cli args because this application uses optparse.

camviewer_ui_impl.py
```diff
@@ -126,7 +126,7 @@ def __getattr__(self, name):
 
 
 class FilterObject(QObject):
```
tangkong, line 128: Maybe it's just me, but I found the fact that the filter installs itself on `FilterObject.__init__` really disorienting (rather than `GraphicUserInterface` instantiating then installing the filter)

I think it's less clear this way, but as they say: "when in Rome"

camviewer_ui_impl.py
```diff
@@ -142,6 +142,8 @@ def __init__(self, app, main):
         self.last = QPoint(0, 0)
 
     def eventFilter(self, obj, event):
+        if event.type() in [QEvent.MouseButtonPress, QEvent.KeyPress]:
+            self.main.refresh_disco()
```
tangkong: 🪩 ?

does this stand for disconnect? (I understand what it does from further reading, but the name struck me)

camviewer_ui_impl.py
```diff
@@ -1297,6 +1307,44 @@ def onSizeUpdate(self):
...
+        self.last_des_max_rate = rate
+        self.set_max_image_rate(rate)
+
+    def set_max_image_rate(self, rate: int) -> None:
+        """
+        Update the maximum allowed rate by restarting the appropriate timer.
+
+        Rate is expected to be a positive integer in Hz.
+
+        When this timer expires, wantImage is called.
```
tangkong: Is this true?  looks like only stop/setDisco are called

camviewer_ui_impl.py
```diff
@@ -1297,6 +1307,44 @@ def onSizeUpdate(self):
...
+        The image rate will determine the rate limiting timeout.
+        1 Hz will never time out
+        5 Hz or less will time out in a week
+        30 Hz and greater will time out in a day
+        """
+        rate = int(rate)
+        if rate <= 0:
+            raise ValueError("Rate must be greater than zero!")
+        self.imageTimer.start(int(1000.0 / rate))
+        self.ui.label_max_rate_value.setText(f"{rate} Hz")
```
tangkong, line 1339: When I check this out my ui file doesn't have this attribute

<details><summary>Details</summary>
<p>

```python
Traceback (most recent call last):
  File "/cds/home/r/roberttk/src/camviewer/./camviewer.pyw", line 94, in <module>
    gui = GraphicUserInterface(
  File "/cds/home/r/roberttk/src/camviewer/camviewer_ui_impl.py", line 511, in __init__
    self.user_set_max_image_rate(rate)
  File "/cds/home/r/roberttk/src/camviewer/camviewer_ui_impl.py", line 1319, in user_set_max_image_rate
    self.set_max_image_rate(rate)
  File "/cds/home/r/roberttk/src/camviewer/camviewer_ui_impl.py", line 1339, in set_max_image_rate
    self.ui.label_max_rate_value.setText(f"{rate} Hz")
AttributeError: 'Ui_MainWindow' object has no attribute 'label_max_rate_value'
```

</p>
</details>

camviewer_ui_impl.py
```diff
@@ -2456,31 +2503,76 @@ def onRemAvgEnter(self):
...
     def stop_disco(self):
         self.discoTimer.stop()
+        self.update_display_timer()
+
+    def refresh_disco(self):
+        if self.last_des_max_rate > 1:
+            self.discoTimer.start()
+            self.set_max_image_rate(self.last_des_max_rate)
 
     def setDisco(self, secs):
```
tangkong: A bit of disambiguation for myself (which I'd argue should be doc-stringed)
- do_disco: sets the app to its "idle state"
  - sets the max image rate to 1, stops the discoTimer and updates display (calls stop_disco)
- stop_disco: stops disco timer, updates display 
- refresh_disco: re-starts discoTimer

When DiscoTimer elapses we run `do_disco`, which sets the app into an idle (less frequently refreshing) state

camviewer_ui_impl.py
```diff
@@ -1297,6 +1307,44 @@ def onSizeUpdate(self):
...
+        if rate <= 0:
+            raise ValueError("Rate must be greater than zero!")
+        self.imageTimer.start(int(1000.0 / rate))
+        self.ui.label_max_rate_value.setText(f"{rate} Hz")
+
+        if rate <= 1:
+            self.stop_disco()
+        else:
+            self.setDisco(
+                np.interp(rate, [5, 30], [self.max_timeout, self.min_timeout])
```
tangkong, line 1345: So between 5 and 30, we interpolate between the max and min timeouts we have set, and outside of that we are capped at the max or min timeouts.  

A nifty little shortcut 👍

camviewer_ui_impl.py
tangkong: This seems to do what you detail on the PR description, I just wanted to make sure I understood the guts.

tangkong (commented):
~~Aside from this not opening for me on local checkout (I assume I'm just missing a ui file update),~~

 I think my biggest gripe is that `{do,stop,refresh}_disco` are a set of unintuitive method names.

camviewer_ui_impl.py
ZLLentz: Yeah, I didn't really want to refactor this more than I needed to

camviewer_ui_impl.py
ZLLentz: Is it worth fixing all of these annoying names? It wouldn't be so hard.

camviewer_ui_impl.py
ZLLentz: `camviewer` doesn't use the ui file at runtime, it expects you to `make` an equivalent `.py` file

camviewer_ui_impl.py
ZLLentz: I can install some docstrings for the next unfortunate maintainer

camviewer_ui_impl.py
ZLLentz: `wantImage` is called when `imageTimer` times out, it's set up somewhere else.

camviewer_ui_impl.py
ZLLentz: I will fix the annoying names that are related to this PR:
- The ones that are too terse
- The ones that use camelCase

ZLLentz: Also:
- The ones whos names make no sense now that I'm not disconnecting any cameras!

I think I was too tepid with renames on the first pass because I was afraid of breaking random undocumented internals.

tangkong: I think future us would appreciate it, in the whole "leave it better than you found it" frame of mind

camviewer_ui_impl.py
tangkong: Ah ok I got mixed up here.  The QTimers aren't blocking, and `wantImage` is connected to `ImageTimer`.  stop/setDisco are called without waiting for this timer

camviewer_ui_impl.py
tangkong: oh no I forgot to make 😮‍💨

camviewer_ui_impl.py
ZLLentz: No more panic at the disco, I renamed nearly everything related to this PR to be much closer to what it actually does

camviewer.pyw
```diff
@@ -76,6 +78,17 @@ if __name__ == "__main__":
     if options.scale is not None:
         os.environ["QT_SCALE_FACTOR"] = options.scale
 
+    if options.min_timeout is None:
+        # 1 day
+        min_timeout = 24 * 60 * 60
```
KaushikMalapati, line 83: Are you intending for users to use min/max_timeout (as opposed to these existing for debugging)? Are we hoping that anyone who does use them won't set mix_timeout really high and understands that we are using units of seconds?

camviewer.ui
```diff
@@ -625,6 +663,81 @@ Vmin\</string>
...
+           </property>
+           <property name="text">
+            <string>NO</string>
+           </property>
+          </widget>
+         </item>
+         <item row="4" column="0" colspan="2">
+          <widget class="QLabel" name="label_max_rate_text">
+           <property name="text">
+            <string>Max Display Rate</string>
```
KaushikMalapati, line 682: The addition of this and the "Rate Timeout In" row make my gui look squished. Does it look normal on a control room monitor? If I am only seeing this because of my laptop's small screen it isn't a problem. 

Before
image

After
image

camviewer_ui_impl.py
```diff
@@ -142,6 +142,8 @@ def __init__(self, app, main):
         self.last = QPoint(0, 0)
 
     def eventFilter(self, obj, event):
+        if event.type() in [QEvent.MouseButtonPress, QEvent.KeyPress]:
```
KaushikMalapati, line 145: Is it worht adding QEvent.Enter (move into window) or QEvent.MouseMove if you also self.setMouseTracking(True) (move cursor within window)?

camviewer_ui_impl.py
```diff
@@ -542,7 +554,6 @@ def __init__(
         self.cross2Update.connect(lambda: self.onCrossUpdate(1))
         self.cross3Update.connect(lambda: self.onCrossUpdate(2))
         self.cross4Update.connect(lambda: self.onCrossUpdate(3))
-        self.timeoutExpiry.connect(self.onTimeoutExpiry)
```
KaushikMalapati, line 545: Should we also remove this signal? https://github.com/pcdshub/camviewer/blob/99521d828b092278dc6ef05f9e6b72fc187fc8e5/camviewer_ui_impl.py#L194

KaushikMalapati:
Overall this looks good to me. I'm not sure I understand why anyone would want a camera to be active for up to a week without interaction when we have twelve-hour shifts.

KaushikMalapati:
> pre-existing idle cli arg is now defunct (not sure what to do with it)
I only see this used in et/camviewer and we can either

- Let it do nothing (current soa)
- Remove it (if we think nobody will use, but probably not worth the time)
- This was previously used as the amount of time in between timeoutdialog popups, so we could make it an analog for max_timeout

camviewer.pyw
ZLLentz: I was only intending these as debug tools. I'm also happy to remove them.

camviewer.ui
ZLLentz: Let me check if there is something I can do for small screens. It's still important to be able to run this app remotely.

camviewer_ui_impl.py
ZLLentz: If these don't have performance penalties I'm happy to add them. Let's look into it.

camviewer_ui_impl.py
ZLLentz: Yes, I forgot to remove the signal.

camviewer.ui
tangkong: This worked fine for me through x-forwarding, fwiw

ZLLentz:
> I'm not sure I understand why anyone would want a camera to be active for up to a week without interaction when we have twelve-hour shifts.

The main consumer who would be happy with multi-shift camera views is the laser lab, which may be viewing the same upstream diagnostics across multiple shifts over the course of a week.

camviewer.pyw
ZLLentz: I guess the second intention for these is to allow us to tune the timeout bounds in https://github.com/pcdshub/engineering_tools/blob/master/scripts/camViewer without needing to make a new release.

ZLLentz: I'm not super concerned with end users setting these parameters incorrectly because most people use the wrapper script to launch the app and once you get into the ui it's pretty clear if you've made a mistake with your timeout numbers.

ZLLentz:
I think I'll leave `--idle` alone as a vestigial arg for scripts that still call it, so they can run without crashing. This application actually has a number of "features" like this which are partially still in the code despite being no longer used...

camviewer.ui
ZLLentz: ~I'm worried~ I suspect that there's something silly like a too-small minimum height on that total section of the display. I'll squint at the ui file for a few minutes.

ZLLentz: I can reproduce this by going full-screen on my laptop monitor with the camera settings added to the right hand section.
The ui file has no minimum height sizing on any of the right-side elements. I'm not sure if setting them would matter.
This behavior is present in master too (with a sufficiently small monitor).

Since this is also present in master (exasperated here by the addition of new ui elements) I'm going to add this to the jira and handle it in a separate PR.

ZLLentz: Possibly it will be necessary to add a scrollbar...

camviewer_ui_impl.py
ZLLentz: I gave this a try, but the performance penalty was much, much worse than I expected. I'm not sure if this is intrinsic to qt/pyqt or a byproduct of how this app is architected, but if you turn on mouse tracking and look for `Enter` and/or `MouseMove` the image rate dips almost to zero whenever you exercise these events. I'm going to opt to not include these.

They did refresh the timer as expected fwiw

camviewer_ui_impl.py
ZLLentz: I've removed this signal and two other unused signals

ZLLentz:
Ok, I believe this is stable now! The reviews were positive plus feedback that has either been addressed or has been added to the camviewer jira. It is time to go next!

camviewer_ui_impl.py
tangkong: Probably due to how frequently we receive a mouseMove event I suppose?  If each of those is allowed to trigger the callback, I can imagine that producing a lot of calls.  

I wonder if it'd help performance to also filter based on if we're in the "idle" state.  This is more of a musing, since that'd change the behavior of the app.

tangkong:
was about to stamp.  No objections here 👍

PR #17, opened 2024-11-22
FIX: atomic writes for config/cache files

ZLLentz:
camviewer config files occasionally get corrupted.
This is likely because users can have multiple copies of the GUI open writing at the same time, non-atomically.

This PR cleans up the file writing so that the final file that ends up in the user's home area config is always placed there atomically via the operating system moving the file. The intermediate writes are applied to a temporary file. The final file is not moved unless it is complete (without exceptions).

In addition, the temporary file is managed with a context manager, which was not done before in this library, making it more likely to get cleaned up in the case of unrelated errors.

I moved the entire write block out of the class body and into the module body, despite still needing a reference to the entire gui object, because the function was already deeply nested and adding 2 context managers on top of that sounded like a maintenance hazard. So, instead, I apply the context managers separately and in the module scope to decrease the indentation pain.

This is similar to happi's atomic writes (https://github.com/pcdshub/happi/blob/e29373c8c09d91d642cd2a0cb0a8aed058754c55/happi/backends/json_db.py#L138-L151), except:
- I have no guarantee that the destination file already exists, so I can't copy mode and instead set a sensible mode manually
- I use the operating system's temp directory instead of making a file in the user's home
- I made a little context manager because I need to do this twice

I did not scrutinize the config file writing other than this specific attention to atomic writes. It's pretty clear there are other things to look at but I wanted to keep scope manageable.

Unrelated edits are appeasements of flake8

### DIFF
--- a/camviewer_ui_impl.py
+++ b/camviewer_ui_impl.py
@@ -2711,16 +2711,16 @@ class GraphicUserInterface(QMainWindow):
         self.cfg = None
 
 
-def write_camera_config(self: GraphicUserInterface) -> None:
-    f = open(self.cfgdir + self.cameraBase, "w")
-    f.write("projsize    " + str(self.projsize) + "\n")
-    f.write("viewwidth   " + str(self.viewwidth) + "\n")
-    f.write("viewheight  " + str(self.viewheight) + "\n")
+def write_camera_config(gui: GraphicUserInterface) -> None:
+    f = open(gui.cfgdir + gui.cameraBase, "w")
+    f.write("projsize    " + str(gui.projsize) + "\n")
+    f.write("viewwidth   " + str(gui.viewwidth) + "\n")
+    f.write("viewheight  " + str(gui.viewheight) + "\n")
     f.write("portrait    " + str(int(param.orientation == param.ORIENT90)) + "\n")
     f.write("orientation " + str(param.orientation) + "\n")
-    f.write("autorange   " + str(int(self.ui.checkBoxProjAutoRange.isChecked())) + "\n")
+    f.write("autorange   " + str(int(gui.ui.checkBoxProjAutoRange.isChecked())) + "\n")
     f.write("use_abs     1\n")
-    rz = self.ui.display_image.rectZoom.abs()
+    rz = gui.ui.display_image.rectZoom.abs()
     f.write(
         "rectzoom    "
         + str(rz.x())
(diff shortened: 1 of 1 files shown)


camviewer_ui_impl.py
```diff
@@ -2789,3 +2713,100 @@ def getConfig(self):
...
+            print(f"Error writing {path}: {exc}")
+        else:
+            # File must be closed before we can chmod and move it
+            fd.close()
+            # Set -rw-r--r-- instead of temp file default -rw-------
+            os.chmod(fd.name, 0o644)
+            shutil.move(fd.name, path)
+    # If the tempfile still exists, we should clean it up.
+    if os.path.exists(fd.name):
+        os.remove(fd.name)
```
tangkong, line 2812: Does python's `tempfile` not take care of its own cleanup?  Could we set `delete=True, delete_on_close=False` here so that we can close/move the file, but let the context manager handle deleting it?

tangkong (approved):
Not too much to object to, I think this makes sense.  A quick question about the tempfile handling

camviewer_ui_impl.py
```diff
@@ -2416,89 +2417,8 @@ def setDispSpec(self, v):
...
-                "projconstant " + str(int(self.ui.checkBoxConstant.isChecked())) + "\n"
-            )
-            f.write("projcalib   %g\n" % self.calib)
-            f.write('projcalibPV "%s"\n' % self.calibPVName)
-            f.write('projdisplayFormat "%s"\n' % self.displayFormat)
-
-            f.close()
-            g.close()
+            write_camera_config(self)
+            write_global_config(self)
```
janeliu-slac, line 2425: That looks much cleaner than the original.

janeliu-slac (commented):
Everything looks good to me, going to approve.

camviewer_ui_impl.py
```diff
@@ -2711,16 +2711,16 @@ def getConfig(self):
         self.cfg = None
 
 
-def write_camera_config(self: GraphicUserInterface) -> None:
-    f = open(self.cfgdir + self.cameraBase, "w")
-    f.write("projsize    " + str(self.projsize) + "\n")
-    f.write("viewwidth   " + str(self.viewwidth) + "\n")
-    f.write("viewheight  " + str(self.viewheight) + "\n")
+def write_camera_config(gui: GraphicUserInterface) -> None:
```
janeliu-slac, line 2718: Just for my own understanding--why was self changed to gui?

camviewer_ui_impl.py
ZLLentz: `delete_on_close` was added in python 3.12 :fire:

janeliu-slac (approved):

camviewer_ui_impl.py
ZLLentz: In `python` typically `self` is used for bound class methods (by convention, not by necessity). Since this isn't a bound class method, I decided to rename the variable so people won't get confused.

PR #19, opened 2024-11-26
WIP: use the count to break early instead of segfault

ZLLentz:
https://jira.slac.stanford.edu/browse/ECS-6914

Notes to future me about past me:
- There's at least one error here in the implementation
- There are more efficient ways to do this, such as checking the pointer location or cutting early via multiplying width x height and checking it against count
- The most important things here for the Jira are preventing the segfault and telling the user what is wrong with their camera config


ZLLentz:
I've decided to implement this by finding the maximum number of rows we can fill given the buffer count and setting the end parameters for the nested loop lower if the count is low. This will be simpler than what's in the PR diff here.

Separately, I want to check apparent count vs pixel count and let the user know if there is a discrepancy.

ZLLentz:
closed in favor of #22

PR #20, opened 2024-11-27
ENH: replace custom EVR launcher with standard expert launcher

ZLLentz:
There's a button in a subscreen in camviewer labelled "Open EVR". There are a lot of issues associated with this button:
- Requires the user to include an EVR PV in the config file
- If no EVR PV is given, this button will still exist but couldn't work
- The button doesn't work anyway, having not been functional since the transition of operator consoles to rhel7
- The config file parsing is bugged and ends up assigning previous EVR PVs to later cameras if they have an empty string
- The script that opens the screen is brittle and would need to be periodically updated

Instead, as per https://jira.slac.stanford.edu/browse/ECS-6881, this PR replaces this button with one that opens the expert screen.

The expert screen is not a special script that needs to live inside this repo. Rather, the IOC defines its own expert screen launcher script and stashes it in the "LAUNCH_GUI" PV (or, the "LAUNCH_EDM" PV for older IOCs, the update has not been rolled out widely).

This PR makes the viewer connect to these PVs on camera setup and decode/stash their values. When a user presses the expert screen button, the script is called as-is if it isn't an empty string. "LAUNCH_GUI" takes priority over "LAUNCH_EDM" if both exist.


camviewer_ui_impl.py
```diff
@@ -2810,3 +2880,21 @@ def atomic_writer(path: str) -> typing.Iterator[typing.TextIO]:
...
+    numpy=True as a kwarg.
+
+    The waveform is an array of signed 8-bit integers whose
+    unsigned representations correspond to the ascii character codes.
+    The string is null-terminated.
+    """
+    # Implementation lifted from PyDM's "parse_value_for_display"
+    zeros = np.where(waveform == 0)[0]
+    if zeros.size > 0:
+        waveform = waveform[: zeros[0]]
```
tangkong, line 2899: an opportunity for my own learning, are there non-zero characters after the first zero?  If it's just zero padding I'm sure there are simpler ways to do this.  (not that I will object much to a mostly proven implementation)

tangkong (approved):
This seems unobjectionable to me.  👍

camviewer_ui_impl.py
ZLLentz: I'll be honest, I didn't explore improving this snippit at all.

I think since the EPICS string is null-terminated, any data after the first 0 is invalid by definition.

tangkong: Right, I was just wondering if we get like: 32321300000 or 3213210321321.  I guess it doesn't matter 🤷

ZLLentz: Yeah, I think in general we expect something like

123 120 142 0 0 0 0 0 0

And we should be able to treat data after the first 0 as "garbage" by definition

But I'm not sure if we ever get these sorts of strings in practice

camviewer_ui_impl.py
```diff
@@ -7,6 +7,8 @@
 # So, we only have to deal with true screen coordinates and how the oriented image is mapped to
 # this.
 #
+from __future__ import annotations
```
KaushikMalapati, line 10: Where do you use an annotation that needs this?

camviewer_ui_impl.py
```diff
@@ -24,10 +26,12 @@
 import time
 import functools
 import numpy as np
+import numpy.typing as npt
 import tempfile
 import shutil
 import typing
 import contextlib
+import subprocess
```
KaushikMalapati, line 34: Would it be simpler if showexpert was a PyDMShellCommand instead of a QPushButton? This way we could get rid of the two script variables, and have a single new_gui_script method that just sets the commnds property to whatever value they receive.

KaushikMalapati:
I left some comments but I think this is good to merge as-is.

camviewer_ui_impl.py
tangkong: I don't think pydm is currently a dependency of camviewer.  It's probably fine to add it as a dependency, but any change to requirements should be looked at carefully.

camviewer_ui_impl.py
tangkong: the `type1 | type2` union syntax was introduced in 3.10, for one ([PEP604](https://peps.python.org/pep-0604/)).  (we still support 3.9 everywhere)

camviewer_ui_impl.py
```diff
@@ -2230,14 +2255,59 @@ def setCalibPV(self, pvname):
     def onSpecific(self, button):
         pass
 
-    def onOpenEvr(self):
-        iCamera = self.ui.comboBoxCamera.currentIndex()
-        if self.lEvrList[iCamera] != "None":
-            print(
-                "Open Evr %s for camera [%d] %s..."
-                % (self.lEvrList[iCamera], iCamera, self.lCameraList[iCamera])
+    def new_launch_gui_script(self, exc: Exception | None = None):
```
KaushikMalapati, line 2258: new_launch_gui_script and new_launch_edm script are very similar. Is it worth passing itertools.partial objects as the monitor arguments with the script and pv variables as arguments?

camviewer_ui_impl.py
ZLLentz: I think completely replacing camviewer with a pydm app would make so many things easier. This is a really good example.

camviewer_ui_impl.py
ZLLentz: Yeah, this is a judgement call for sure. I ultimately decided that having it copied once was easier than formulating the partials because I was thinking about some old article I read about practical coding that insisted that copying once was ok and copying twice was the greatest sin. In retrospect I don't know how much I agree with that article.

Some of these camviewer PRs are a bit rough in that I didn't put in quite as much care as I would with one of our Python libraries, mostly because I'm trying to put camviewer to bed to work on the replacement app. I should probably slow down a bit.

camviewer_ui_impl.py
ZLLentz: Regardless, hooking up the button to a bash one-liner with caget would have definitely been more practical to implement than what I ended up doing here, so I probably should have considered it.

camviewer_ui_impl.py
KaushikMalapati: > Some of these camviewer PRs are a bit rough in that I didn't put in quite as much care as I would with one of our Python libraries, mostly because I'm trying to put camviewer to bed to work on the replacement app. I should probably slow down a bit.

I think copying it is probably fine then. The main reason to do this would be so you don't have to duplicate edits if anyone ever updated new_launch_XXX_script, but that's probably unlikely

tangkong: I think there's a reasonable middle ground to be found between writing verbose/readable code and playing code golf.  

If camviewer 2.0 is on the horizon, I think it's more than reasonable to button up these quick fixes and move on.

ZLLentz: Ok, I'll merge this one and spend my attention on the more topical things you both pointed out in the image-saving PR.

If we need to change these again I'll refactor at that time.

camviewer_ui_impl.py
KaushikMalapati: In case anyone is curious, something like this might have worked (not asking for an update)
`f"$(caget -tS {self.ctrlBase}:LAUNCH_GUI 2&> /dev/null || caget -tS {self.ctrlBase}:LAUNCH_EDM)"`

ZLLentz: Yes, exactly.

The only weird edge case where this PR's implementation is "better" is when the IOC goes up and down often and you want to open the expert screen when the IOC is "down", in this case we'd have already cached the script name and wouldn't fail a caget.

PR #21, opened 2024-11-27
ENH: report errors, retry in saving camera image

ZLLentz:
Related jira: https://jira.slac.stanford.edu/browse/ECS-6918

This PR's primary purpose is to fix a bug where the image saving dialog always reports success.

The reason for this is that `QImage.save` never raises, instead it returns `True` or `False` to signal if the image was saved properly. The original author of this bit of code assumed that the save function would raise with information about what went wrong.

This ends up being somewhat annoying, leaving it to us to figure out why the save might have failed.

Changes implemented here:

- Correctly check if `QImage.save` succeeded or failed
- If it failed, do a best-effort check for simple issues
- If it failed, ask the user to try a different filename and re-open the dialog
- Set the default directory to the user's home area, which is guaranteed to be writeable, unlike the current working directory
- fileName -> filename because the naming was irritating me

### DIFF
--- a/camviewer_ui_impl.py
+++ b/camviewer_ui_impl.py
@@ -1177,8 +1177,13 @@ class GraphicUserInterface(QMainWindow):
             # QImage.save returned False, so the save failed.
             # Check for obvious errors, then retry the save
             # No write permissions?
-            directory = os.path.dirname(filename)
-            if not os.access(path=directory, mode=os.W_OK):
+            # If the file exists, we need to check the filename. Otherwise, we check the dirname.
+            if os.path.exists(filename):
+                can_write = os.access(path=filename, mode=os.W_OK)
+            else:
+                directory = os.path.dirname(filename)
+                can_write = os.access(path=directory, mode=os.W_OK)
+            if not can_write:
                 return self.warn_and_retry_save(
                     message=f"No permissions to write {filename}! Please pick a different location."
                 )
--- a/camviewer_ui_impl.py
+++ b/camviewer_ui_impl.py
@@ -1183,8 +1183,8 @@ class GraphicUserInterface(QMainWindow):
                     message=f"No permissions to write {filename}! Please pick a different location."
                 )
             # Invalid image type?
-            image_types = ["npy"] + [
-                qba.data().decode("utf-8")
+            image_types = [".npy"] + [
+                "." + qba.data().decode("utf-8")
                 for qba in QImageWriter.supportedImageFormats()
             ]
             if file_ext not in image_types:
--- a/camviewer_ui_impl.py
+++ b/camviewer_ui_impl.py
@@ -1151,30 +1151,56 @@ class GraphicUserInterface(QMainWindow):
 
     def onfileSave(self):
         try:
-            fileName = QFileDialog.getSaveFileName(
+            filename = QFileDialog.getSaveFileName(
                 self,
                 "Save Image...",
                 self.cwd,
                 "Images (*.npy *.jpg *.png *.bmp *.pgm *.tif)",
             )[0]
-            if fileName == "":
+            if filename == "":
                 raise Exception("No File Name Specified")
-            if fileName.lower().endswith(".npy"):
-                np.save(fileName, self.image)
+            if filename.lower().endswith(".npy"):
+                np.save(filename, self.image)
                 QMessageBox.information(
                     self,
                     "File Save Succeeded",
-                    "Image has been saved as a numpy file: %s" % (fileName),
+                    "Image has been saved as a numpy file: %s" % (filename),
                 )
-                print("Saved to a numpy file %s" % (fileName))
+                print("Saved to a numpy file %s" % (filename))
+            elif self.ui.display_image.image.save(filename, format=None, quality=-1):
+                # QImage.save returned True, so the save succeeded.
+                QMessageBox.information(
+                    self,
+                    "File Save Succeeded",
+                    "Image has been saved to an image file: %s" % (filename),
+                )
+                print("Saved to an image file %s" % (filename))
             else:
-                self.ui.display_image.image.save(fileName, format=None, quality=-1)
-                QMessageBox.information(
+                # QImage.save returned False, so the save failed.
+                # Check for obvious errors, then retry the save
+                # File already exists?
+                if os.path.exists(filename):
+                    reason = (
+                        f"{filename} already exists! Please pick a different filename"
+                    )
+                # No write permissions?
+                else:
+                    # This is the most robust way to check if we have permissions
+                    # os.access exists but doesn't quite work fully on NFS filesystems as per the Python docs
+                    try:
+                        with open(filename, "w"):
+                            ...
+                    except PermissionError:
+                        reason = f"No permissions to write to {filename}! Please pick a different location."
+                    else:
+                        # Clean up the stub file we made
+                        os.remove(filename)
+                QMessageBox.warning(
                     self,
-                    "File Save Succeeded",
-                    "Image has been saved to an image file: %s" % (fileName),
+                    "File Save Failed",
+                    reason,
                 )
-                print("Saved to an image file %s" % (fileName))
+                return self.onfileSave()
         except Exception as e:
             print("fileSave failed:", e)
             QMessageBox.warning(self, "File Save Failed", str(e))
--- a/camviewer_ui_impl.py
+++ b/camviewer_ui_impl.py
@@ -43,6 +43,7 @@ from PyQt5.QtGui import (
     QClipboard,
     QPixmap,
     QDrag,
+    QImageWriter,
 )
 from PyQt5.QtCore import (
     QTimer,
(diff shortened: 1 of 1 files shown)


camviewer_ui_impl.py
```diff
@@ -17,6 +17,7 @@
 
 import sys
 import os
+import os.path
```
KaushikMalapati: Why did you import os.path if you already had os?

ZLLentz: I was under the mistaken impression that `import os` did not expose the utilities in `os.path`, because this is how all (most) other modules work. I'll remove the redundant import. I suspect this is a habit I picked up by reading code from other programmers.

There's some interesting implementation details about this I was unaware of, namely that the `os` module imports either `posixpath` or `ntpath` based on your OS and re-exports it as `os.path`, injecting it into `sys.modules` along the way. Pretty wild behavior, but I guess this is a reasonable way to aim for platform-independent libraries.

tangkong: The `os.path` platform specifics is a TIL for me as well

I'm pretty sure I'd have just tried to access the `path` as a submodule of `os`, but Zach is right that this is generally not suggested.  This access method only works if the parent module imports its submodules.  This isn't guaranteed to work for all modules nor is it guaranteed to not change.

camviewer_ui_impl.py
```diff
@@ -1151,30 +1151,58 @@ def shutdown(self):
...
                     self,
                     "File Save Succeeded",
-                    "Image has been saved as a numpy file: %s" % (fileName),
+                    "Image has been saved as a numpy file: %s" % (filename),
                 )
-                print("Saved to a numpy file %s" % (fileName))
-            else:
-                self.ui.display_image.image.save(fileName, format=None, quality=-1)
+                print("Saved to a numpy file %s" % (filename))
+            elif self.ui.display_image.image.save(filename, format=None, quality=-1):
```
tangkong: It's weird to me to see expressions with side-effects in conditional arguments.  I think it's a code smell thing primarily, but it's also readable enough.  

Refactoring would probably just be pulling the `image.save` call out and starting a new `if else` block

camviewer_ui_impl.py
```diff
@@ -1151,30 +1151,58 @@ def shutdown(self):
...
+                # Check for obvious errors, then retry the save
+                # File already exists? This shouldn't happen due to the qt file dialog.
+                if os.path.exists(filename):
+                    reason = (
+                        f"{filename} already exists! Please pick a different filename"
+                    )
+                # No write permissions?
+                else:
+                    # This is the most robust way to check if we have permissions
+                    # os.access exists but doesn't quite work fully on NFS filesystems as per the Python docs
```
tangkong: I've seen `os.stat` mentioned in conjunction with the `stat` module, if there's a need for an alternate way?  ([stackoverflow](https://stackoverflow.com/questions/1861836/checking-file-permissions-in-linux-with-python))

camviewer_ui_impl.py
```diff
@@ -1151,30 +1151,58 @@ def shutdown(self):
...
+                        os.remove(filename)
+                        # I guess we have no idea what went wrong
+                        reason = "Unknown failure! Please try again."
+                QMessageBox.warning(
+                    self,
+                    "File Save Failed",
+                    reason,
                 )
-                print("Saved to an image file %s" % (fileName))
+                return self.onfileSave()
```
tangkong: are we recursing on purpose here?

camviewer_ui_impl.py
ZLLentz: I agree, I probably shouldn't have combined these together, if only for readability. It's pretty easy to get confused reading this as-is.

camviewer_ui_impl.py
ZLLentz: I think `os.stat` also won't fully encompass the NFS permissions but maybe the accuracy drop will be worth the simpler implementation (not accidentally creating a stub file)

camviewer_ui_impl.py
ZLLentz: Restarting the procedure is intended, the recursive call is the easiest way to do this.
Probably re-emitting a signal is the cleaner way.

camviewer_ui_impl.py
```diff
@@ -1151,30 +1151,58 @@ def shutdown(self):
...
-            fileName = QFileDialog.getSaveFileName(
+            filename = QFileDialog.getSaveFileName(
                 self,
                 "Save Image...",
-                self.cwd,
+                os.path.expanduser("~"),
                 "Images (*.npy *.jpg *.png *.bmp *.pgm *.tif)",
             )[0]
-            if fileName == "":
+            if filename == "":
```
KaushikMalapati, line 1163: Is it intentional to exit if there's no filename provided but retry if there was a saving problem?

camviewer_ui_impl.py
```diff
@@ -1151,30 +1151,58 @@ def shutdown(self):
...
                 raise Exception("No File Name Specified")
-            if fileName.lower().endswith(".npy"):
-                np.save(fileName, self.image)
+            if filename.lower().endswith(".npy"):
+                np.save(filename, self.image)
```
KaushikMalapati, line 1169: Just double checking, but can this fail silently? In the cases you check for if QImage.save fails (no write permission, file already exists, other), can this fail by throwing an exception which would not result in a retry? Also looking at https://github.com/numpy/numpy/blob/v2.1.0/numpy/lib/_npyio_impl.py#L507-L588
I think that this could override an existing file with the same name which shouldn't be possible in case someone tries to use the same filename for different images and ends up overwriting all but the last one.

camviewer_ui_impl.py
```diff
@@ -1151,30 +1151,58 @@ def shutdown(self):
 
     def onfileSave(self):
         try:
-            fileName = QFileDialog.getSaveFileName(
+            filename = QFileDialog.getSaveFileName(
```
KaushikMalapati, line 1157: If you clicked on the "Save to File" button and then decided not to save anything by exiting the dialog, what happens here? Does getSaveFileName return nothing causing an exception?

camviewer_ui_impl.py
```diff
@@ -1151,30 +1151,58 @@ def shutdown(self):
...
                     "File Save Succeeded",
-                    "Image has been saved to an image file: %s" % (fileName),
+                    "Image has been saved to an image file: %s" % (filename),
+                )
+                print("Saved to an image file %s" % (filename))
+            else:
+                # QImage.save returned False, so the save failed.
+                # Check for obvious errors, then retry the save
+                # File already exists? This shouldn't happen due to the qt file dialog.
+                if os.path.exists(filename):
```
KaushikMalapati: Can QImage make the fail but still return False?

camviewer_ui_impl.py
janeliu-slac: Yep, agree that side effects in conditional statements aren't ideal. (I had another comment about moving a print statement--please disregard--I misread the code)

camviewer_ui_impl.py
ZLLentz: It's not documented very well (https://doc.qt.io/qt-5/qfiledialog.html) but empty string is returned if the user clicks "cancel".

camviewer_ui_impl.py
ZLLentz: You're right that I need to be more careful here. The exception being thrown would be "safe" but wouldn't result in a retry, giving inconsistent behavior.

I'll double-check about re-using the same name, but my fuzzy memory is that the file dialog gives you an "are you sure?" sort of error, so if we reach `np.save` the user has already confirmed that they want to replace the file.

camviewer_ui_impl.py
KaushikMalapati: Is this meant for this comment https://github.com/pcdshub/camviewer/pull/21#discussion_r1866794027?

camviewer_ui_impl.py
ZLLentz: It ends up returning an empty string. Qt APIs can be awkward like that.

camviewer_ui_impl.py
ZLLentz: The API docs claims that it cannot, but it's good to be skeptical, especially with pyqt as a middleman between us and the qt docs. Maybe I should also simplify here and be less ambitious in checking some of these things.

https://doc.qt.io/qt-5/qimage.html#save

camviewer_ui_impl.py
ZLLentz: No, I meant to respond here in this way, but maybe I didn't justify myself fully.

The intention is to exit if the user clicks "cancel", which is signaled by there being no filename. Otherwise the user has no way to change their mind and not save anything.

KaushikMalapati: Feels slightly wrong to throw an exception and have a warning popup if you press cancel, but not sure if it's worth making more user-friendly

ZLLentz: Ah yeah, you're right- there's a UX problem here. I forgot what came next when responding to this comment since it was pre-existing code. I'll clean this up.

camviewer_ui_impl.py
ZLLentz: Resolved by refactoring. It should now be clear how this is being used.

camviewer_ui_impl.py
ZLLentz: Resolved via simplifying with `os.access` to reduce compexity

camviewer_ui_impl.py
ZLLentz: Resolved via swapping out recursive call with a signal

camviewer_ui_impl.py
ZLLentz: Resolved by making this code path give no message at all. No raise, no pop-up- the UI element that gets us to this point doesn't let you input an empty string, so the only way to be here is to click "X" or "Cancel", so you don't need this confirmation pop-up in any form. Added a comment to clarify intent.

camviewer_ui_impl.py
ZLLentz: Resolved by refactoring to make specifically `OSError` call the same method that we reach via the `QImage.save` checks.
Confirmed that the file select dialog warns the user sufficiently if they pick an already existing filename.

camviewer_ui_impl.py
ZLLentz: Resolved by removing the exception as stated above

camviewer_ui_impl.py
ZLLentz: Resolved by abandoning this check in favor of large simplification. This at least avoids an incorrect message in the case where QImage would have undefined behavior, and it allows the user to freely override existing filenames if they want to for whatever reason.

ZLLentz:
In addition to the above, I added a check for invalid file types since it's the only other situation I could find where QImage.save would fail aside from permissions issues. There's a limited, predefined set of image types that are valid.

ZLLentz:
The diff is somewhat confusing now because it's more like a rewrite than a tweak, but I think this is cleaner than before!

camviewer_ui_impl.py
tangkong: Interestingly python suggests the ask-for-forgiveness (try-catch) approach for security reasons we probably don't care about.  

Out of curiosity, do you remember what the NFS limitations were?

ZLLentz: I don't remember the details, but in network file systems there can be additional file protections that go beyond the posix uid/guid stuff. I think in practice this isn't used for us.

camviewer_ui_impl.py
```diff
@@ -1151,33 +1154,70 @@ def shutdown(self):
...
+                filename, format=None, quality=-1
+            )
+            if save_ok:
+                # QImage.save returned True, so the save succeeded.
+                return self.show_file_success(filename=filename, file_ext=file_ext)
+            # QImage.save returned False, so the save failed.
+            # Check for obvious errors, then retry the save
+            # No write permissions?
+            directory = os.path.dirname(filename)
+            if not os.access(path=directory, mode=os.W_OK):
```
tangkong: If the file exists, should we check access permissions on the chosen file itself, rather than the directory?  Currently if I try to write to a file that exists but I have no permissions on, this will report an invalid image type...  (I wrote a file, write-protected it, then tried to save to that same file again)
image

This also reveals a subtle bug where `image_types` has extensions without the `.`, but `file_ext` includes `.` (e.g. `.jpg`)

ZLLentz: Good catch. I originally was checking file permissions but that gave `False` if the file didn't exist. I'll check for existence and check the file or the folder as appropriate.

I'll clean up the user-facing reporting to make it more clear these are file extensions, and I'll make sure the ext variable matches the list so that we can reach the "unknown error" thing without having false positives on the extension check.

ZLLentz: I'm starting to regret not standing up unit tests for camviewer, as annoying as that would have been.

KaushikMalapati:
@ZLLentz how do I run this locally after cloning your branch?

tangkong:
I had to:
* be on a rhel7 server
* `make all`
* Copy a camviewer config from a `xxxopr` home directory
* run `run_viewer_local.sh` with appropriate options to select a camera
  * e.g.: `./run_viewer_local.sh --instrument XCS --pvlist camviewer.cfg --camera 12`

tangkong (approved):
This looks good to me now.  I think it's worth pushing forward with this, and moving on to greater things.

KaushikMalapati (approved):
I think this is good. It's possible to save a file with only the extension but it's probably not worth checking for.

PR #22, opened 2024-12-18
FIX: set max row count to avoid iterating off the end of the data buffer

ZLLentz:
## For reviewers

I'm not quite ready to deploy this but I'd like opinions from people who have experience writing C++.
I'd particularly like tips or ideas for doing this more intelligently.

## Description

There's a rare bug that deals with improperly-configured cameras. I suspect it also could happen as a race condition on a camera resize.

This bug manifests itself when the "count" of the epics array is shorter than the total number of pixels. In some cases this results in garbage data making it into the image, in other cases this results in a segfault.

This update aims to add minimal processing overhead to the new array processor while stopping it from iterating past the end of the data buffer. It does this by using the previously unused "count" long from pyca which tells us how many data elements there are, and doing a quick one-time calculation of the max row we can get to without exhausting the buffer. Then we sub this number in for the row count in the iteration and stop right after the final row.

This works in my toy example I cooked up and it doesn't crash the GUI, but before merging I'd need to intentionally misconfigure a camera to check what the behavior ends up being.

### DIFF
--- a/camviewer_ui_impl.py
+++ b/camviewer_ui_impl.py
@@ -1991,6 +1991,22 @@ class GraphicUserInterface(QMainWindow):
         # Get camera configuration
         self.getConfig()
 
+        # Check the expected size against the count to generate warnings
+        first_image_count = len(self.camera.value)
+        expected_count = self.rowPv.value * self.colPv.value
+        if first_image_count != expected_count:
+            QMessageBox.warning(
+                None,
+                "Warning",
+                (
+                    "IOC misconfiguration likely!\n"
+                    f"Received {first_image_count} pixels, expected {expected_count} "
+                    f"for {self.rowPv.value} x {self.colPv.value}."
+                ),
+                QMessageBox.Ok,
+                QMessageBox.Ok,
+            )
+
     def setCameraMenu(self, index):
         for a in self.camactions:
             a.setChecked(False)
--- a/pycaqtimage/pycaqtimage.sip
+++ b/pycaqtimage/pycaqtimage.sip
@@ -349,8 +349,17 @@ static int rowIncMult1[8] = {  0,   0,  -1,  -1,   0,   0,   1,   1}; /* Scaled 
 static int rowIncMult2[8] = {  0,   2,   0,   0,   0,  -2,   0,   0}; /* Scaled by srcwidth */
 static int rowIncK[8]     = {  0,   0,  -1,   1,   0,   0,   1,  -1}; /* Constant */
 
+int _max_row(ImageBuffer *imageBuffer, long count)
+{
+    if (imageBuffer->imgwidth > 0) {
+        return std::min(imageBuffer->imgheight, (int) std::floor(count / imageBuffer->imgwidth));
+    } else {
+        return 0;
+    }
+}
+
 template <class T>
-void _pyDoAvg(ImageBuffer *imageBuffer, T *cadata)
+void _pyDoAvg(ImageBuffer *imageBuffer, T *cadata, long count)
 {
     T *src           = cadata;
     uint32_t *dst    = imageBuffer->imageData;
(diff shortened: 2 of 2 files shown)


pycaqtimage/pycaqtimage.sip
```diff
@@ -349,8 +349,17 @@ static int rowIncMult1[8] = {  0,   0,  -1,  -1,   0,   0,   1,   1}; /* Scaled
...
 
+int _max_row(ImageBuffer *imageBuffer, long count)
+{
+    if (imageBuffer->imgwidth > 0) {
+        return std::min(imageBuffer->imgheight, (int) std::floor(count / imageBuffer->imgwidth));
+    } else {
+        return 0;
+    }
+}
+
```
janeliu-slac, line 360: This is minor but if using a modern C++ approach I think a smart pointer is considered best practice. Raw pointers should be used only when performance/memory management are critical.

If only one pointer for ImageBuffer is needed, std::unique_ptr would work.
```
#include <memory> 

int _max_row(unique_ptr<ImageBuffer> imageBuffer, long count)
{
    if (imageBuffer->imgwidth > 0) {
        return std::min(imageBuffer->imgheight, (int) std::floor(count / imageBuffer->imgwidth));
    } else {
        return 0;
    }
}
```
Regarding the buffer overflow problem--if the program uses `std::vector`, `std::array` or other dynamically sized containers similar to a python list maybe this issue can be avoided. Raw arrays is probably one of the major reasons the world moved away from C.

baljamal:
I like your solutions. in addition I have suggests that I think it will limit segfualt problems:

1) using -fsanitize compile flag to make it easier to detect such problems
2) replace c-style arrays with std::array with exception handling

baljamal (commented):
I like your solutions. in addition I have suggestions that I think it will limit segfualt problems:

1. using -fsanitize compile flag to make it easier to detect such problems
2. replace c-style arrays with std::array with exception handling

janeliu-slac (dismissed):

ZLLentz:
Thank you Basil and Jane- I'll use your comments and guidance as a starting point for some C++ learning for me in the new year. If we can make this less likely to crash with bad inputs it's definitely worthwhile.

camviewer_ui_impl.py
```diff
@@ -1991,6 +1991,22 @@ def connectCamera(self, sCameraPv, index, sNotifyPv=None):
         # Get camera configuration
         self.getConfig()
 
+        # Check the expected size against the count to generate warnings
+        first_image_count = len(self.camera.value)
```
ZLLentz, line 1995: Note to future me: does this work with color cameras? Need to check

ZLLentz:
I looked through the docs of the various things mentioned here and some of them are very applicable while others are hard to apply because of the API we need to match from pyca/EPICS.

In the interest of practicality, what I'm going to do is clean this PR up with the current implementation (I think I just need to make sure this doesn't break color cams, which have 3x as many array values), then I'll create follow-up issues for the code health improvements mentioned here.

ZLLentz:
It turns out that this segfaults on orientation changes so it needs some more thought

ZLLentz:
Basically, the image renderer does something wild: it does not iterate through the incoming data buffer pointer in order. Instead, it iterates through the data buffer pointer in some wild order depending on the orientation of the image. It fills in the pixels left-to-right top-to-bottom but it might iterate through the pointer backwards, or sometimes it jumps around a lot etc. based on the image rotation/flip state.

This means that "max row count" is an ineffective guard on this bad behavior.

I'm going to see if there's any way to salvage this but it's likely that I'll simplify and just prevent this code from being called at all if there is an issue.

ZLLentz:
I'm going to dump this altogether and make a new PR. I'm going to log the best-practices suggestions as a github issue.

janeliu-slac:
Hmm sounds like the image renderer should go through a separate function that forces it to iterate through the data buffer pointer in the right sequence. Could there be a race condition involved as well (where the image renderer reads then fills the pixels)? If so I've used std::atomic in the past to force a read and write operation to be treated as one atomic operation.

ZLLentz:
> sounds like the image renderer should go through a separate function that forces it to iterate through the data buffer pointer in the right sequence

The data buffer is already being iterated through in some wild but correct sequence. Separating it into a function seems correct to me but I don't have a strong reason to prioritize this.

> Could there be a race condition involved as well

I think there is never a race condition due to the careful (but insane) way the data is iterated through. It reads each pixel in the source data pointer and writes to each pixel in the image data pointer exactly once.

PR #23, opened 2024-12-19
ENH: better behavior regarding "offline" cams and rate indicators

ZLLentz:
## For reviewers

I'm mostly interested in thoughts on general code qc and implementations that seem odd (either wrong or needing further explanation).
The easiest way to test this PR is:
- psbuild-rhel7
- clone, checkout
- make
- `./run_viewer_local.sh --instrument las --pvlist /cds/group/pcds/pyps/config/las/camviewer.cfg`

`las` has a few cameras at different rates and a few totally offline cameras, so it's a good place to use.

This PR increases the rate of on-close segfaults because it increases the number of ca monitors. I'll opt to clean these up properly in a follow-up PR . I don't want so many core dumps being sent around but I also think this problem predates the PR here a bit.

Note that camviewer core dumps go to the /tmp directory...

## Problem statement
https://jira.slac.stanford.edu/browse/ECS-6882

In short, without this PR, if you select a camera with an offline IOC, you get hit with a 3x 5s timeout penalty.

## Major Changes
- In the cameras menu (top left), cameras will indicate whether they are normal (no extra text), stopped (`"(Stopped)"` text), or if the ioc is offline (`"(Offline)"` text). This is not replicated in the top right combobox because there isn't enough room.
- Now that the app knows if a camera is online, picking an offline camera is a quick error instead of the previous 3x 5s timeouts with user acknowledgement.
- Now that the app is always connected to all camera's rate PVs, this info is now used to display a "Cam rate", replacing the old "Data rate" indicator.
- In the distant past, "Data rate" and "Display rate" were different numbers, but these days they are always the same number. So, dropping one is not a loss of information. It would have been possible to fix this instead of replace it but the value I picked is both easier to get and more useful, and putting something like 4 different rates on the screen is a bit of a waste.
- The application only supports area detector cameras now, but there was a lot of code leftover from when it supported various LCLS-only cameras. This PR also removes a lot of this vestigial code because some small part of it was getting in my way.
- Minor cleanup


ZLLentz:
TIL I could just do `Pv.value` instead of `Pv.data["value"]` in `psp`/`pyca`, I'll simplify these

camviewer.ui
```diff
@@ -697,6 +700,9 @@ Vmin\</string>
            <property name="text">
             <string>-</string>
            </property>
+           <property name="alignment">
+            <set>Qt::AlignLeading|Qt::AlignLeft|Qt::AlignVCenter</set>
```
KaushikMalapati, line 704: Shouldn't this also be AlignRight?

ZLLentz: It looked really silly at `AlignRight`, so I switched it back to `AlignLeft`

tangkong:
this worked nicely for me, I was able to click through cameras and the offline popups were quick and painless.
I was left wishing there was a slider to adjust the width of the toolbar for long camera names
image

Or possibly make the combo box options expand via `SizeAdjustPolicy`?

Edit: seems like the containing widgets have very strict size policies that make this difficult.  I won't object to this too loudly, that's a ui rework for another day.  

Edit 2: It's also less necessary with the menu bar holding camera names, which I didn't notice before

ZLLentz:
I notice the cam rate is clipping in your screenshot too. Maybe there are a number of layout adjustments that could be done for this app...

tangkong (approved):
This looks good to me.  The deletions make sense, and the app works nicely for me 👍

ZLLentz:
I'm going to merge this _after_ the end of year tag. I want to get the follow-up PR in that removes the on-close core dumps before deploying this in the wild.

PR #24, opened 2024-12-20
FIX: fix error in recolor logic and simplify

ZLLentz:
There is a sip-wrapped C++ function that the GUI calls when it needs to recolor an image that was already rendered. It does this, for example, when the user changes the color map or adjusts the min/max pixel value sliders so that stale images are recolored live.

There was a bug reported that this doesn't actually work (https://jira.slac.stanford.edu/browse/ECS-7015)
and upon further investigation, the logic was wrong.

The original code had a comment indicating that the recolor should be skipped for color cams and when averaging was on.
Then, it had an if statement that did exactly the opposite: recolor would only be done for color cams and when averaging was on.

I ended up fixing and then simplifying the check to only exclude color cams. Including averages is useful because when you average e.g. 100 frames it can take a long time before you get the next image, so it's nice to see the colors update live. Excluding color cams is important because it makes the screen flash black.

### DIFF
--- a/pycaqtimage/pycaqtimage.sip
+++ b/pycaqtimage/pycaqtimage.sip
@@ -329,9 +329,10 @@ static void _pyCopyToQImage(ImageBuffer* imageBuffer, int doFC)
 PyObject* pyRecolorImageBuffer(PyObject* pyImageBuffer)
 {
     ImageBuffer* imageBuffer = (ImageBuffer*) PyCapsule_GetPointer(pyImageBuffer, PYC_IB);
-    /* If the image is supposed to be color or if we're averaging, skip it! */
-    if ((imageBuffer->isColor && !imageBuffer->useGray) || imageBuffer->iAverage != 1)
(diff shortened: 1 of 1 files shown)


tangkong (approved):
seems to do what it says on the tin 👍 This also avoided the core-dump-on-close in the one time I tried it

KaushikMalapati (commented):
> Including averages is useful because when you average e.g. 100 frames it can take a long time before you get the next image, so it's nice to see the colors update live.

If you average something in color 1 and then swap to color 2, is it the same as averaging in color 2 to begin with?

ZLLentz:
> If you average something in color 1 and then swap to color 2, is it the same as averaging in color 2 to begin with?

The average is done prior to applying the color map. So, we accumulate N images, average them, store those averaged values as the apparent raw image pixel values, and only then do we remap these into colors.

So, swapping to color map 2 or changing the min/max re-uses the final result of the averaging calculation against a new color mapping rule.

KaushikMalapati (approved):

PR #25, opened 2024-12-20
ENH: auto range for min/max sliders

ZLLentz:
Part of https://jira.slac.stanford.edu/browse/ECS-7005

Main changes:
- Calculate the min/max pixel in the ROI on each frame. I added this to the C++ code in the same block where we calculate other statistics about the ROI.
- Add button to set the min/max of the color map range to these values. This could be used to narrow the range once before further manual tuning of the color range.
- Add checkbox to automatically update the min/max color map ranges on each frame. This could be used to continuously adjust the color range for changing input intensities.

Related internal changes:
- Rework the various range widgets to only apply their changes when the user causes a change. This allows us to have separate dedicated internal functions for updating the related internal attributes and another set of functions for updating the UI display of these values that doesn't cause an infinite loop.

image

### DIFF
--- a/camviewer.ui
+++ b/camviewer.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>1119</width>
-    <height>1075</height>
+    <width>1106</width>
+    <height>1070</height>
    </rect>
   </property>
   <property name="sizePolicy">
@@ -798,38 +798,6 @@ Vmin\</string>
          <string>Color Map</string>
         </property>
         <layout class="QGridLayout" name="gridLayout_4">
-         <item row="0" column="0" colspan="2">
-          <widget class="QComboBox" name="comboBoxColor">
-           <property name="currentIndex">
-            <number>1</number>
-           </property>
-           <item>
-            <property name="text">
-             <string>HSV</string>
-            </property>
-           </item>
-           <item>
-            <property name="text">
-             <string>Hot</string>
-            </property>
-           </item>
-           <item>
-            <property name="text">
-             <string>Jet</string>
-            </property>
-           </item>
-           <item>
-            <property name="text">
-             <string>Cool</string>
-            </property>
-           </item>
-           <item>
-            <property name="text">
-             <string>Gray</string>
-            </property>
-           </item>
-          </widget>
-         </item>
          <item row="0" column="3">
           <widget class="QComboBox" name="comboBoxScale">
            <item>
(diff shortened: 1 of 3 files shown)


ZLLentz:
I'm also going to ask UED if they like the verbiage/layout here or if something else is preferred.

I've been testing this using:
```
make
```

camviewer.ui
```diff
@@ -898,6 +898,30 @@ Vmin\</string>
            </property>
           </widget>
          </item>
+         <item row="2" column="3">
+          <widget class="QLineEdit" name="lineEditRangeMax">
```
tangkong, line 902: Any reason we shouldn't make this a QSpinBox?

tangkong (commented):
Somewhat tragic that qt doesn't have a range slider built-in.   A quick search turned up some homegrown solutions but I don't think that's worth the effort here.

Code here looks mostly good to me.  My main concerns (shifting color map ranges when refreshing automatically, self referential loops when moving slider and updating values) seem to have been taken care of, but I'll take the ui for a spin to verify

camviewer.ui
ZLLentz: No reason, a QSpinBox would be strictly better. It would let us constrain the min/max values more precisely, allow us to stop doing bonus str <-> int conversions ourselves, and has built-in input validation. I can swap it over.

ZLLentz: I think this is also a rare case where I might like the spinbox arrows. Putting the ranges up or down in a few clicks could be nice.

janeliu-slac:
Hmm I tried to run `make` on camviewer but it it's not finding Qt/qimage.h
```
pycaqtimage.sip:7:10: fatal error: Qt/qimage.h: No such file or directory
    7 | #include <Qt/qimage.h>
```
echo $PATH shows
`/reg/g/pcds/epics/extensions/R1.4.0/bin/rhel9-x86_64:/reg/g/pcds/epics/base/R7.0.2-2.0/bin/rhel9-x86_64:/cds/sw/tools/bin:/cds/group/pcds/pyps/conda/py39/envs/pcds-5.9.1/epics/bin/linux-x86_64:/cds/group/pcds/pyps/conda/py39/envs/pcds-5.9.1/bin:/cds/group/pcds/pyps/conda/py39/condabin:/reg/g/pcds/engineering_tools/latest-released/scripts:/cds/group/pcds/pyps/apps/fzf/bin:/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/reg/g/pcds/plc-common/InstallFiles/modpoll/linux:/reg/g/pcds/plc-common/InstallFiles/modpoll/linux:/cds/home/j/janeliu/.local/bin:/cds/home/j/janeliu/bin`

ZLLentz:
Unfortunately it doesn't build on rocky9 yet
When I fixed the rocky9 build I ended up breaking the rhel7 build so I set it aside for now

janeliu-slac (approved):
Got it compiling and running. LGTM.

tangkong (approved):
Seems to work as advertised for me.  The one suggestion I have is that it might be useful to also have the sliders scale.  Right now if your max/min are <50 or so, the slider's workable area is very narrow.  

But also that is a lot of work so feel free to ignore that suggestion.

ZLLentz:
I'll give the spinbox idea a go but I'm really burnt out on sliders

ZLLentz:
For ease of reviewing, that spinbox will be a follow-up PR

PR #26, opened 2024-12-20
FIX: close monitors at app exit to avoid shutdown core dumps

ZLLentz:
After #23, where I added a lot of monitors, there were core dumps at shutdown most times I closed the app.

When you inspect these core dumps, you'd see that what's going on is that the app is terminating during the process of a pyca epics callback, raising an exception in the c++ code, and then dropping a core dump into `/tmp`.

This PR collects all the monitored PVs during the close event and stops all the monitors, guaranteeing that the monitor callbacks will not be running when the application fully closes and the PV objects are garbage collected.

To test, I used xcs this time because they have a bunch of online cams (and someone turned off all the laser IOCs)

```
```

Without the patch, I opened and closed the app 5 times and got 5 core dumps.
With the patch, I opened and closed the app ~15 times and got 0 core dumps. 

Ref https://jira.slac.stanford.edu/browse/ECS-7016

### DIFF
--- a/camviewer_ui_impl.py
+++ b/camviewer_ui_impl.py
@@ -654,6 +654,7 @@ class GraphicUserInterface(QMainWindow):
         self.efilter = FilterObject(self.app, self)
 
     def closeEvent(self, event):
+        self.end_monitors()
         if self.cameraBase != "":
             self.activeClear()
         if self.haveforce and self.forcedialog is not None:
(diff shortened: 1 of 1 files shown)


camviewer_ui_impl.py
```diff
@@ -665,6 +666,24 @@ def closeEvent(self, event):
...
+        all_mons.extend(self.otherpvs)
+        all_mons.append(self.notify)
+        all_mons.append(self.rowPv)
+        all_mons.append(self.colPv)
+        all_mons.append(self.launch_gui_pv)
+        all_mons.append(self.launch_edm_pv)
+        all_mons.append(self.lensPv)
+        all_mons.append(self.calibPV)
+        for pv in all_mons:
+            if pv is not None:
```
tangkong, line 683: do pyca pvs delete themselves and replace themselves with `None` when they're disconnected?

tangkong (approved):
This looks mostly fine, can't really argue with properly tearing down.

camviewer_ui_impl.py
ZLLentz: This is a camviewer special where it stuffs `None` as a stand-in for these attributes when they fail to connect

PR #27, opened 2025-01-10
ENH: convert text edits to spinboxes for integer color scaling limits

ZLLentz:
Follow-up to #25 to make the input widgets for the integer color map limits use `QSpinBox` instead of `QLineEdit`.

`QSpinBox` is the most correct native widget to get basic `int` inputs from the user, as opposed to `QLineEdit` which natively outputs strings.

I had to tweak various bits of the code but overall it's simpler and more functional like this.

### DIFF
--- a/camviewer.ui
+++ b/camviewer.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>1106</width>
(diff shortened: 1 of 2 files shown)


tangkong (approved):
LGTM 👍

PR #28, opened 2025-01-10
ENH: cam-specific controls

ZLLentz:
Request from UED
Awaiting feedback in https://jira.slac.stanford.edu/browse/ECS-7005 but I think the code is done now.

The main idea here is that UED wanted camera-specific controls for their Andor cameras because they need to change the gain a lot and using two GUIs is burdensome to operations.

Every cam will get a camera controls section on the main screen with at least start/stop controls, and specific camera models can be granted extra behavior.

This is implemented by adding an additional empty QGroupBox, giving it a QFormLayout, and repeatedly emptying and filling this layout as needed when cameras swap around.

I implemented this in a submodule and had it clean itself up on camera switches instead of trying to weave it into the big camviewer file. I didn't use any ui files because I wanted to keep this simple and limited.

image

### DIFF
--- a/camviewer_ui_impl.py
+++ b/camviewer_ui_impl.py
@@ -522,7 +522,7 @@ class GraphicUserInterface(QMainWindow):
         self.refresh_timeout_display_timer.setInterval(1000 * 20)
         self.refresh_timeout_display_timer.start()
 
-        self.model_screen_generator = None
+        self.cam_type_screen_generator = None
         self.setup_model_specific()
 
         self.ui.average.returnPressed.connect(self.onAverageSet)
@@ -672,8 +672,8 @@ class GraphicUserInterface(QMainWindow):
         self.end_monitors()
         if self.cameraBase != "":
             self.activeClear()
-        if self.model_screen_generator is not None:
-            self.model_screen_generator.cleanup()
+        if self.cam_type_screen_generator is not None:
+            self.cam_type_screen_generator.cleanup()
         if self.haveforce and self.forcedialog is not None:
             self.forcedialog.close()
         self.advdialog.close()
@@ -1138,8 +1138,8 @@ class GraphicUserInterface(QMainWindow):
                 pv.disconnect()
             except Exception:
                 pass
-        if self.model_screen_generator is not None:
-            self.model_screen_generator.cleanup()
+        if self.cam_type_screen_generator is not None:
+            self.cam_type_screen_generator.cleanup()
         self.otherpvs = []
 
     def shutdown(self):
@@ -2038,16 +2038,16 @@ class GraphicUserInterface(QMainWindow):
         self.getConfig()
 
     def setup_model_specific(self):
-        if self.model_screen_generator is None:
+        if self.cam_type_screen_generator is None:
             form = QFormLayout()
             self.ui.groupBoxControls.setLayout(form)
         else:
-            self.model_screen_generator.cleanup()
+            self.cam_type_screen_generator.cleanup()
             form = self.ui.groupBoxControls.layout()
-        self.model_screen_generator = CamTypeScreenGenerator(self.ctrlBase, form)
-        self.model_screen_generator.final_name.connect(self.new_model_name)
-        if self.model_screen_generator.full_name:
-            self.new_model_name(self.model_screen_generator.full_name)
+        self.cam_type_screen_generator = CamTypeScreenGenerator(self.ctrlBase, form)
+        self.cam_type_screen_generator.final_name.connect(self.new_model_name)
+        if self.cam_type_screen_generator.full_name:
+            self.new_model_name(self.cam_type_screen_generator.full_name)
 
     def new_model_name(self, name: str):
         self.ui.groupBoxControls.setTitle(f"{name} Controls")
--- a/camviewer_ui_impl.py
+++ b/camviewer_ui_impl.py
@@ -1138,9 +1138,10 @@ class GraphicUserInterface(QMainWindow):
                 pv.disconnect()
             except Exception:
                 pass
-        if self.cam_type_screen_generator is not None:
-            self.cam_type_screen_generator.cleanup()
         self.otherpvs = []
+        self.cameraBase = ""
+        self.ctrlBase = ""
+        self.setup_model_specific()
 
     def shutdown(self):
         self.clear()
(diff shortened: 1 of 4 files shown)


models.py
tangkong, line 1: Perhaps a dangerously overloaded term in the model-view-controller world qt lives in.  But maybe that's just my biases talking

models.py
```diff
@@ -0,0 +1,205 @@
...
+            self.model = self.model_pv.value
+            self.model_ready.emit()
+
+    def finish_form(self) -> QFormLayout:
+        if not self.manufacturer or not self.model:
+            return
+        with self.finish_lock:
+            if self.finish_ran:
+                return
+            self.finish_ran = True
```
tangkong: If we encounter an error anywhere in the body of this, we'll be locked out from ever being able to `.finish_form()` again.  Is this what you intended?  I suppose we're ok with this since people aren't really interacting with these methods

models.py
```diff
@@ -0,0 +1,205 @@
...
+    gain_set_pv.connect()
+
+    def set_gain_value():
+        try:
+            gain_set_pv.put(gain_spinbox.value())
+        except Exception:
+            ...
+
+    gain_spinbox = QSpinBox()
+    gain_spinbox.setMaximum(10000000)
```
tangkong: that's a lot of gain

tangkong (commented):
This seems reasonable to me, I'll take it for a spin before stamping.  

Almost done with camviewer huh? 😆

tangkong:
Apparently the UED camera IOCs are down so I can't test these on the UED cameras in question, but I do think it's nice that we can see a more fully qualified name for the camera in camviewer now.

ZLLentz:
UED cam IOCs are only available from the UED subnet, not even to psbuild.

tangkong:
ah right that special subnet.  Silly me

models.py
ZLLentz: You are absolutely right here. This should be renamed to something less overloaded.
Maybe cam_types.py?

models.py
ZLLentz: I was much more concerned with accidentally running this twice than I was with the form failing to complete.

ZLLentz: Maybe this was not correct

models.py
ZLLentz: I have no idea what the actual correct number is and the jira is pretty silent

tangkong:
I think I managed to break it by selecting a camera that was indeed down, then trying to swiftly switch back to one of the Andors.  

Here I have a gige selected but the andor specific controls are there, and the entire UI is frozen

image

tangkong:
Which server are you testing this on?  I just picked a UED server but maybe I'm still on the wrong subnet

models.py
tangkong: Give them all the gain!  I imagine we should probably be grabbing a max value from the PV metadata, but that can't be trusted 😢

models.py
tangkong: I'm happy with cam_types.py 👍

models.py
ZLLentz: the metadata says the limits are 0 and 0
I could go to the manual I guess

models.py
tangkong: I think it's a valid concern, and one that you addressed rather thoroughly.  Perhaps reading this from the point of view of `ModelScreenGenerator`, I was just inherently distrustful of the downstream functions

ZLLentz:
> Here I have a gige selected but the andor specific controls are there, and the entire UI is frozen

And dangerously, the start/stop buttons are still visible from the previous camera... I'll have to add an additional clear step.

> Which server are you testing this on?

I was using `ued-daq` and being really careful not to actually change their settings. I was also running on `tst-console` with `/cds/home/z/zlentz/camviewer.cfg` with a silly temporary IOC hooked up to a camera in b901.

tangkong:
Interesting, that's where I ended up too  (ued-daq).

Another thing to point out is how the spinbox didn't get initialized to the same value as the gain number?

models.py
tangkong: That's on the IOC creator :/

ZLLentz:
>  the spinbox didn't get initialized to the same value as the gain number?

I forgot to do this, I will add it

models.py
ZLLentz: Fwiw I couldn't find a manual that put a bound on this value

models.py
ZLLentz: I ended up wrapping the downstream functions a bit to make sure they raise a warning pop-up instead of crashing the app

ZLLentz:
I think I addressed the items here but I will spend some time later making sure I didn't compromise the user experience. The way I've implemented the more thorough cleanup is a bit inconsistent, for example.

ZLLentz:
- Responded to review suggestions
- Added docstrings
- Cleaned up the UX for not-loading cameras
- Disabled the run widgets when they wouldn't work (e.g. no cam loaded, no write access)
- Disabled the EM Gain widgets when the cam is running (UED says cam running + gain change = dead cam)

camviewer_ui_impl.py
```diff
@@ -2026,6 +2038,21 @@ def connectCamera(self, sCameraPv, index, sNotifyPv=None):
...
         self.getConfig()
 
+    def setup_model_specific(self):
+        if self.cam_type_screen_generator is None:
+            form = QFormLayout()
+            self.ui.groupBoxControls.setLayout(form)
+        else:
+            self.cam_type_screen_generator.cleanup()
+            form = self.ui.groupBoxControls.layout()
+        self.cam_type_screen_generator = CamTypeScreenGenerator(self.ctrlBase, form)
```
tangkong, line 2048: So every time we set a new camera, we create a new `CamTypeScreenGenerator`?  Do we ever unparent the old ones or deleteLater them? 

In flipping between cameras I never ran into a problem/memory leak, but I am suspicious.  It seems like we're already disconnecting signals and removing rows from the form, so could we just repopulate it?

Ultimately I think I'm ok with this as is.  If we run into performance issues we can revisit.

tangkong (approved):
The problems I ran into in my previous tests have been resolved 🎉 

Looks like things get cleaned up nicely as well.  I'm onboard  👍

camviewer_ui_impl.py
ZLLentz: You are correct, replacing this object should no longer be necessary with some small tweaks. (I won't fix this now, in favor of forward progress through the last few camviewer tickets)

In an earlier version of this PR I was trying to remove the `QFormLayout` from the `QGroupBox` which actually seems to be impossible in pyqt. In that earlier version the screen generator was a one-time use object that had custody of a unique layout per screen generator object, so I was removing and replacing them. It was only later that I switched to clearing and re-using the layout instead.

PR #30, opened 2025-01-15
FIX: wrong image size handling

ZLLentz:
Alternative to #22 

I think this is the simplest possible variant to avoid the crash.
Simply compare the count to the size (which is set as width*height). If they are not the same, warn the user and do not process any image data.

In practice, this only affects us when the IOC is misconfigured. Otherwise the data will always match and we will get good frames.

If the camera is misconfigured, this PR will result in:
- A warning
- A completely black (pixel values =  0) image

### DIFF
--- a/camviewer_ui_impl.py
+++ b/camviewer_ui_impl.py
@@ -2026,6 +2026,24 @@ class GraphicUserInterface(QMainWindow):
         # Get camera configuration
         self.getConfig()
 
+        # Check the expected size against the count to generate warnings
+        first_image_count = len(self.camera.value)
+        expected_count = self.rowPv.value * self.colPv.value
+        if self.isColor:
+            expected_count *= 3
+        if first_image_count != expected_count:
+            QMessageBox.warning(
+                None,
+                "Warning",
+                (
+                    "IOC misconfiguration likely!\n"
+                    f"Received {first_image_count} pixels, expected {expected_count} "
+                    f"for {self.rowPv.value} x {self.colPv.value}."
+                ),
+                QMessageBox.Ok,
+                QMessageBox.Ok,
+            )
+
     def normalize_selectors(self):
         """
         Update the visual appearance of the camera combobox and the menu to be correct.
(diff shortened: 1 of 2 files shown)


janeliu-slac:
I'm curious how the C++ code can be refactored to get the image renderer to read the data buffer pointer in the right sequence but this looks like a good solution in the meantime.

ZLLentz:
I'm also curious about the most correct way to refactor the code such that we get as many properly-placed pixels as possible before running out of data, but also I think forcing people to configure their data streams properly isn't a big ask.

I'm intending to put future efforts into reworking/replacing the image renderer entirely, getting the data via a different protocol and using external libraries to handle things like image orientation for example.

camviewer_ui_impl.py
```diff
@@ -2026,6 +2026,24 @@ def connectCamera(self, sCameraPv, index, sNotifyPv=None):
         # Get camera configuration
         self.getConfig()
 
+        # Check the expected size against the count to generate warnings
+        first_image_count = len(self.camera.value)
+        expected_count = self.rowPv.value * self.colPv.value
+        if self.isColor:
+            expected_count *= 3
```
tangkong, line 2045: Any image that is color is always RGB?   No funny CMYK or otherwise?

tangkong (dismissed):
Seems like what had to be done.

camviewer_ui_impl.py
ZLLentz: I checked into this a little bit, there are a lot of instances of `3` hard-coded into the sip file, so it seems like it. There are direct references to RGB as well.

In general (separate to the source code) I expect CMYK to only be used for printers and for all computer screen graphics to use RGB due to additive vs subtractive color theory.

ZLLentz:
Review was dismissed due to the merge conflict resolution but it's the same PR still.
Merging and heading into the "home stretch" of "final camviewer tickets"

camviewer_ui_impl.py
tangkong: I'll admit I am not an expert on color theory, I just saw a magic number and had to question it haha.

PR #31, opened 2025-01-16
ENH/FIX: remaining small tweaks and verbiage clarification

ZLLentz:
This PR intends to clear all of the remaining items from the camviewer sprint and some additional minor items:

See related jira tickets:
- https://jira.slac.stanford.edu/browse/ECS-6748
- https://jira.slac.stanford.edu/browse/ECS-7014
- https://jira.slac.stanford.edu/browse/ECS-7022
- https://jira.slac.stanford.edu/browse/ECS-7066

Implemented here:
- Wrap the entire right-hand column in a scroll area to facilitate running camviewer on small screens. Do some event filtering to make the sizing feel good.
- Slightly edit the projection widget so it doesn't need to be instantiated after some of the other ui elements, since we have little control over the uic loading order. The order changed a lot when I implemented the scroll bar.
- Set up the expert screen PVs first, so that if something else fails later we can reach the expert screen.
- Copy the expert screen button to the main screen.
- Set the new expert button to be enabled only when the PV is ready for it to work.
- Include a text indicator of which area detector image stream we are viewing because I didn't want the expert button to be the only thing in the box and also because I always wonder which stream I'm viewing when doing debug.
- Clear the image if we connect but find a 0x0 sized image. Otherwise, we end up viewing a stale image because no new frames come through.
- Replace the Connected: YES/NO text with more detailed status information, and make it update live as the connected IOC goes up and down.
- Fix minor sizing issues

image
image

### DIFF
--- a/camviewer_ui_impl.py
+++ b/camviewer_ui_impl.py
@@ -2030,6 +2030,9 @@ class GraphicUserInterface(QMainWindow):
             or self.colPv is None
             or self.colPv.value == 0
         ):
+            # Clear the image so that we don't have a stale image from the previous cam
+            self.ui.display_image.image.fill(0)
+            self.after_new_min_or_max_pixel()
             self.ui.label_connected.setText("NO")
             return
 
--- a/camviewer_ui_impl.py
+++ b/camviewer_ui_impl.py
@@ -567,6 +567,7 @@ class GraphicUserInterface(QMainWindow):
         self.ui.showspecific.triggered.connect(self.doShowSpecific)
         self.ui.actionGlobalMarkers.triggered.connect(self.onGlobMarks)
         self.advdialog.ui.showexpert.clicked.connect(self.on_open_expert)
+        self.ui.show_expert_button.clicked.connect(self.on_open_expert)
         self.onExpertMode()
 
         self.ui.checkBoxProjRoi.stateChanged.connect(self.onGenericConfigChange)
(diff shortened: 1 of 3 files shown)


ZLLentz:
This is intended to be last non-bugfix PR of this sprint (hopefully no more big requests come in later...)

ZLLentz:
I thought I was done but the 0x0 clear image thing only sometimes works, tomorrow I'll try to finish. I discovered this while trying to take a screenshot of it working.

tangkong:
I guess text sizing is slightly different for me, the cam rate label clips a bit.  Interestingly, since the right-hand control panel can't be resized horizontally, this will always be the case.
image

Aside from this the ui seems to work well.  I'll take a peek at the code now.

camviewer_ui_impl.py
```diff
@@ -176,6 +176,30 @@ def eventFilter(self, obj, event):
         return False
 
 
+class ScrollSizeFilter(QObject):
+    """
+    Adjust the width of the scroll area when the scrollbar is shown or hidden.
+
+    This improves the UX by avoiding having the scrollbar overlapping any
+    other ui components.
```
tangkong, line 184: Musing:  This is probably also fixed by allowing the button panel to vary in width, either via type hints or a splitter.  I'm not suggesting we implement this here, but the fixed width seems to have caused problems in more than one place.

ZLLentz:
This was happening to Patrick too... I wonder what the right way to deal with this is.

tangkong (approved):
Things look good to me, nothing out of the ordinary to me in the code.

ZLLentz:
I'm going to handle the 0x0 bug and something for the text sizing in the next PR
I think it might be time to add a splitter

PR #32, opened 2025-01-17
FIX: actually clear the image when we load an invalid cam

ZLLentz:
A follow-up to #31 to fix how we need to clear the live image when we select a cam that nominally connects but doesn't have any pixel values in the image (and therefore, nothing to render).

In these cases we want to clear the image so it's clear that we aren't still looking at the previous camera.

This is how I originally had it (working), but then I decided to try it another way, and both seemed to work, so I went along with the second variant, but my new way never actually worked at all and it was just NFS propagation being slow that fooled me into thinking both ways were equivalent. I had unknowingly only tested it with this variant.

Tested via running in xcs, picking a working cam, and then switching to e.g. xcs_gige_3
```
```

### DIFF
--- a/camviewer_ui_impl.py
+++ b/camviewer_ui_impl.py
@@ -2051,7 +2051,7 @@ class GraphicUserInterface(QMainWindow):
         ):
             # Clear the image so that we don't have a stale image from the previous cam
             self.ui.display_image.image.fill(0)
-            self.after_new_min_or_max_pixel()
+            self.ui.display_image.repaint()
             # Report which issue we had
             if self.rowPv is None or self.colPv is None:
                 self.ui.label_status.setText("IOC timeout in setup")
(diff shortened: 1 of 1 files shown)


tangkong:
I guess my clicking around wasn't thorough enough, sorry.  

Even with this patch, if I click from working camera -> xcs_gige_3, I just get stuck with the last frame of the previous camera

(Or at least all the live xcs cams I picked were mostly low intensity, producing a black image.  And the x/y projections didn't clear)

tangkong:
If the python file is updated this change should take effect right?

ZLLentz:
It should just need the one line python file change
xcs_gige_8 -> xcs_gige_3 is pretty dramatic
any of the black/white cams are dramatic if you click the auto scale button first

ZLLentz:
could we be getting platform-dependent or x-forward-method-dependent behavior?

tangkong:
iirc patrick is on a mac...

tangkong:
If I go from xcs_gige_1 to xcs_gige_3, the image goes to black, but I can click autoscale and get the image back  🤔 

image

ZLLentz:
I set out to disable the scaling buttons when the image is 0x0 but then I thought I should extend this behavior. Now, the scaling widgets will disable whenever they don't do anything, for example if no camera is selected or if we're viewing a color camera in color mode.

ZLLentz:
I'm intending to include this in a tag this week

tangkong (approved):
This looks good to me, both code wise and in interactive testing.  👍

PR #33, opened 2025-01-18
FIX: right side font sizing

ZLLentz:
Somewhat experimental: dynamically adjust the size of the right hand area based on the apparent rendered font size. I set the width to be about 50 characters because this matches my locally observed size that I've been designing on.

The goal is for people running on different laptops which normally have their text get cut off to have more normal sizing behavior.

I have no idea if this actually works. I'll need to try it on a few different monitors. I'll need to bug Patrick too. Something for next week.

Incidentally:
- Changed all the group boxes from fixed to preferred sizing so they can resize to fill the scroll widget. This means I don't have to have the resize logic manually resize every single widget at startup.
- Fixed a scroll resizing race condition that stemmed from installing the event filter too early.

Alternatives explored:
- Using a splitter was tried, but it was too terrible to implement without exploding the layout in the camviewer.ui file
- Setting the widget widths to be something other than "fixed" caused the group boxes to explode out the right size of the window...

### DIFF
--- a/camviewer_ui_impl.py
+++ b/camviewer_ui_impl.py
@@ -50,6 +50,7 @@ from PyQt5.QtGui import (
     QPixmap,
     QDrag,
     QImageWriter,
+    QFontMetricsF,
 )
 from PyQt5.QtCore import (
     QTimer,
@@ -692,6 +693,16 @@ class GraphicUserInterface(QMainWindow):
             self.onCameraSelect(int(cameraIndex))
         except Exception:
             pass
+
+        # Set the right hand area's width based on font sizes
+        font = self.ui.labelCamera.font()
+        fm = QFontMetricsF(font)
+        charwidth = fm.averageCharWidth()
+        # We want to make it wide enough for 50 average chars
+        width = 50 * charwidth
+        self.ui.rightScrollArea.setFixedWidth(width)
+        self.ui.rightPanelWidget.setFixedWidth(width)
+
         self.efilter = FilterObject(self.app, self)
         self.scroll_size_filter = ScrollSizeFilter(self)
 
--- a/camviewer_ui_impl.py
+++ b/camviewer_ui_impl.py
@@ -190,14 +190,25 @@ class ScrollSizeFilter(QObject):
         self.base_size = gui.ui.rightPanelWidget.width()
         self.scroll_area = gui.ui.rightScrollArea
         self.scroll_area.verticalScrollBar().installEventFilter(self)
+        # Start in the correct state
+        if self.scroll_area.verticalScrollBar().isVisible():
+            self.set_yes_scroll_size()
+        else:
+            self.set_no_scroll_size()
+
+    def set_yes_scroll_size(self):
+        self.scroll_area.setFixedWidth(
+            self.base_size + self.scroll_area.verticalScrollBar().width()
+        )
+
+    def set_no_scroll_size(self):
+        self.scroll_area.setFixedWidth(self.base_size)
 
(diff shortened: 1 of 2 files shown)


ZLLentz:
I'm intending to test this on a few endstation computers and on a few laptops that had width troubles prior to this PR
This may or may not be on my list to get merged depending on the results

ZLLentz:
In TMO on the big monitors, this PR fixes the clipped text.

In RIX and locally for me this PR has no obvious affect (they are the same size and just as readable with no clipping)

I'm curious what it does on other laptops

tangkong:
image
This creates a much wider right-hand-side-bar for me and seems to avoid all the clipping issues I saw before 🎉 

(macos, x-forwarding, off psbuild)

camviewer_ui_impl.py
```diff
@@ -694,7 +706,28 @@ def __init__(
...
+        # Timer with 0 delay = end of event queue
+        # i.e. immediately after all the queued initial rendering
+        self.late_init_timer = QTimer(self)
+        self.late_init_timer.singleShot(0, self.late_init)
+
+    def late_init(self):
+        """
+        Init routines to be done immediately after the initial render
+        """
+        # Avoid weird sizing race conditions from doing this in __init__
```
tangkong, line 730: sigh qt really does end up being a bunch of spaghetti if we let it huh... 

I guess the problems arise from setting this filter then having the scroll area size change throughout init?

camviewer_ui_impl.py
```diff
@@ -694,7 +706,28 @@ def __init__(
             self.onCameraSelect(int(cameraIndex))
         except Exception:
             pass
+
+        # Set the right hand area's width based on font sizes
+        font = self.ui.labelCamera.font()
+        fm = QFontMetricsF(font)
+        charwidth = fm.averageCharWidth()
+        # We want to make it wide enough for 50 average chars
+        width = 50 * charwidth
```
tangkong, line 715: Did this come from any specific calculation?  Or was it just an estimation?  (seems good either way)

tangkong (approved):
Works for me on my machine, so it must be good everywhere else!

camviewer_ui_impl.py
ZLLentz: This is worse or stranger than anything you were expecting

When I did this during init, I saw the following sequence at startup:
- show event filter starts
- hide event filter starts
- hide event filter ends
- show event filter ends

So even though the sequence was show first, then hide, so we should have the smaller of the two width values, we ended with the show width value because the event handler executions were interleaved... resulting in a scrollbar-sized extra empty space to the right.

Delaying till later skips these first few events without writing a janky "skip your first event ok please" sort of deal into the event filter, which seemed like it could be fragile.

camviewer_ui_impl.py
ZLLentz: The fixed width predating all of this camviewer work was 300 and it looked good on my monitor at 300. My monitor's charwidth was tenths or hundredths of a pixel over 6, and 50*6=300, so I picked 50 to match my existing sizing and hoped that this ratio would carry forward to other displays.

ZLLentz:
It's worked on 4/4 places tried and I'm confident it doesn't crash anything so let's gooooo

camviewer_ui_impl.py
tangkong: I agree with the approach taken here, I'd rather apply the event filter late than have to track the ins-and-outs of the initial event sequence

tangkong:
Also R.I.P. the splitter dream.

PR #34, opened 2025-02-01
ENH: rework global marker handling

ZLLentz:
See https://jira.slac.stanford.edu/browse/ECS-7128 for the full motivation and context.

### Main Goals
- Make "global" marker mode use all four markers instead of only the first two markers.
- Allow seamless transitions between local and global marker modes.
- Make it really clear which markers are global and which are local via marking them with G or L.
- Completely separate the saved local marker positions from the global marker positions, so that neither set overrides the other.
- Allow for "partial" loading of global markers in the event that only some of the PVs exist.

#### Notable Internal Fixes / Adjustments
- Allow us to cleanly swap markers in and out in DisplayImage to help separate global and local marker position data in memory.
- Remove vestigial code for an older version that had weird special global 2 mode for markers 3, 4 with different behavior.
- Improve the performance of the global marker code by removing an unnecessary dumpConfig step, which suddenly was slowing down the GUI when I doubled the number of global markers.
- Rework the config loading code to remove strange code branches and give us consistent first-load behavior for new cameras.

image

### DIFF
--- a/DisplayImage.py
+++ b/DisplayImage.py
@@ -458,20 +458,63 @@ class DisplayImage(QWidget):
             self.gui.dumpConfig()
 
     def set_one_marker(self, index: int, marker: param.Point):
+        """
+        Replace any of the marker cross locations with a new Point object.
+
+        This allows us to swap marker locations in and out as necessary in
+        other parts of the application.
+
+        The Point object can be retained by the caller and its coordinates
+        can be changed to update the marker position here.
+
+        Parameters
+        ----------
+        index : int
+            Whether to replace marker 0, 1, 2, or 3
+            which correspond to markers 1, 2, 3, and 4 on the GUI.
+        marker : Point
+            An object that stores the x, y location of the associated marker.
+        """
         self.lMarker[index] = marker
 
     def set_markers(self, markers: list[param.Point]):
+        """
+        Replace all of the marker cross locations with a list of Point objects.
+
+        This is a shortcut for calling set_one_marker four times on each of the
+        valid marker indices.
+
+        Parameters
+        ----------
+        markers: list of Point
+            A list of 4 objects that store the x, y locations of their
+            associated markers.
+        """
         for index in range(4):
             self.set_one_marker(index, markers[index])
 
 
 def default_markers() -> list[param.Point]:
+    """
+    Return a list of 4 Points that represent the default marker positions.
+
+    The default marker positions are 100 points horizontally and 100 points
+    vertically away from each of the corners of the image, not on the image
+    itself but off of it.
+    """
     pts = [param.Point(0, 0), param.Point(0, 0), param.Point(0, 0), param.Point(0, 0)]
     reset_markers(pts)
     return pts
 
 
 def reset_markers(markers: list[param.Point]) -> None:
+    """
+    Given a list of 4 Points that represent marker positions, reset them to the defaults.
+
+    The default marker positions are 100 points horizontally and 100 points
+    vertically away from each of the corners of the image, not on the image
+    itself but off of it.
+    """
     markers[0].setAbs(-100, 100)
     markers[1].setAbs(param.x + 100, -100)
     markers[2].setAbs(param.x + 100, param.y + 100)
--- a/DisplayImage.py
+++ b/DisplayImage.py
@@ -457,8 +457,12 @@ class DisplayImage(QWidget):
         if self.gui.cfg is None:
             self.gui.dumpConfig()
 
+    def set_one_marker(self, index: int, marker: param.Point):
+        self.lMarker[index] = marker
+
     def set_markers(self, markers: list[param.Point]):
-        self.lMarker = markers
+        for index in range(4):
+            self.set_one_marker(index, markers[index])
 
 
 def default_markers() -> list[param.Point]:
--- a/DisplayImage.py
+++ b/DisplayImage.py
@@ -515,7 +515,7 @@ def reset_markers(markers: list[param.Point]) -> None:
     vertically away from each of the corners of the image, not on the image
     itself but off of it.
     """
-    markers[0].setAbs(-100, 100)
+    markers[0].setAbs(-100, -100)
     markers[1].setAbs(param.x + 100, -100)
     markers[2].setAbs(param.x + 100, param.y + 100)
     markers[3].setAbs(-100, param.y + 100)
--- a/DisplayImage.py
+++ b/DisplayImage.py
@@ -33,12 +33,7 @@ class DisplayImage(QWidget):
         self.sWindowTitle = "Camera: None"
 
         # set marker data
-        self.lMarker = [
-            param.Point(-100, -100),  # image
-            param.Point(param.x + 100, -100),
-            param.Point(param.x + 100, param.y + 100),
-            param.Point(-100, param.y + 100),
-        ]
+        self.lMarker = default_markers()
         self.lPenColor = [
             (0 / 255.0, 128 / 255.0, 255 / 255.0),  # matplotlib colors!!
             (255 / 255.0, 0 / 255.0, 0 / 255.0),
@@ -118,12 +113,7 @@ class DisplayImage(QWidget):
         self.gui.updateRoiText()
         self.gui.updateMiscInfo()
         if reset:
-            self.lMarker = [
-                param.Point(-100, -100),  # image
-                param.Point(param.x + 100, -100),
-                param.Point(param.x + 100, param.y + 100),
-                param.Point(-100, param.y + 100),
-            ]
+            reset_markers(self.lMarker)
 
     def pWidth(self):
         if param.orientation & 2:
@@ -466,3 +456,19 @@ class DisplayImage(QWidget):
         self.update()
         if self.gui.cfg is None:
             self.gui.dumpConfig()
+
+    def set_markers(self, markers: list[param.Point]):
+        self.lMarker = markers
+
+
+def default_markers() -> list[param.Point]:
+    pts = [param.Point(0, 0), param.Point(0, 0), param.Point(0, 0), param.Point(0, 0)]
+    reset_markers(pts)
+    return pts
+
+
+def reset_markers(markers: list[param.Point]) -> None:
+    markers[0].setAbs(-100, 100)
+    markers[1].setAbs(param.x + 100, -100)
+    markers[2].setAbs(param.x + 100, param.y + 100)
+    markers[3].setAbs(-100, param.y + 100)
--- a/camviewer_ui_impl.py
+++ b/camviewer_ui_impl.py
@@ -323,6 +323,8 @@ class GraphicUserInterface(QMainWindow):
         self.acquire_image_timer = QTimer()
         self.rate_limit_timer = QTimer()
         self.refresh_timeout_display_timer = QTimer()
+        self.cam_changeable_restore = QTimer()
+        self.glob_changeable_restore = QTimer()
 
         self.ui = Ui_MainWindow()
         self.ui.setupUi(self)
@@ -905,7 +907,17 @@ class GraphicUserInterface(QMainWindow):
         self.ui.local_button.setChecked(not use_global)
         self.setUseGlobalMarkers(use_global, on_init=on_init)
 
-    def setUseGlobalMarkers(self, ugm: bool, on_init: bool =False):
+    def allow_glob_changes(self, allow: bool = True):
+        """
+        Helper function to briefly disable all the global vs local widgets.
+        """
+        self.ui.actionGlobalMarkers.setEnabled(allow)
+        self.ui.global_button.setEnabled(allow)
+        self.ui.local_button.setEnabled(allow)
+        if not allow:
+            self.glob_changeable_restore.singleShot(1000, self.allow_glob_changes)
+
+    def setUseGlobalMarkers(self, ugm: bool, on_init: bool = False):
         """
         Apply global or local mode.
 
@@ -918,9 +930,10 @@ class GraphicUserInterface(QMainWindow):
             Therefore, we only need to take further action if the mode
             is changing. Set this to True while initializing a camera
             to ensure that the new camera's markers replace the
-            previous camera's markers.        
+            previous camera's markers.
         """
         if ugm != self.useglobmarks or on_init:
+            self.allow_glob_changes(False)
             self.useglobmarks = ugm
             if ugm:
                 self.connectMarkerPVs()
@@ -1200,35 +1213,20 @@ class GraphicUserInterface(QMainWindow):
     def clear(self):
         self.ui.label_dispRate.setText("-")
         self.ui.label_status.setText("-")
-        if self.camera is not None:
-            try:
-                self.camera.disconnect()
-            except Exception:
-                pass
-            self.camera = None
-        if self.notify is not None:
-            try:
-                self.notify.disconnect()
-            except Exception:
-                pass
-            self.notify = None
-        if self.lensPv is not None:
-            try:
-                self.lensPv.disconnect()
-            except Exception:
-                pass
-            self.lensPv = None
-        if self.putlensPv is not None:
-            try:
-                self.putlensPv.disconnect()
-            except Exception:
-                pass
-            self.putlensPv = None
+        self.camera = self.disconnectPv(self.camera)
+        self.notify = self.disconnectPv(self.notify)
+        self.nordPv = self.disconnectPv(self.nordPv)
+        self.nelmPv = self.disconnectPv(self.nelmPv)
+        self.rowPv = self.disconnectPv(self.rowPv)
+        self.colPv = self.disconnectPv(self.colPv)
+        self.calibPV = self.disconnectPv(self.calibPV)
+        self.launch_gui_pv = self.disconnectPv(self.launch_gui_pv)
+        self.launch_edm_pv = self.disconnectPv(self.launch_edm_pv)
+        self.lensPv = self.disconnectPv(self.lensPv)
+        self.putlensPv = self.disconnectPv(self.putlensPv)
+        self.disconnectMarkerPVs()
         for pv in self.otherpvs:
-            try:
-                pv.disconnect()
-            except Exception:
-                pass
+            self.disconnectPv(pv)
         self.otherpvs = []
         self.cameraBase = ""
         self.ctrlBase = ""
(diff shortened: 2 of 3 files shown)


ZLLentz:
Clicking around a bit, the first issue I see is that my new button doesn't properly show the status when switching to a new camera. Oops!

The per-marker indicator is always correct because it's triggered specifically by the PV callback routines.

ZLLentz:
One more TODO per conversation in the Jira: I will stop having the global marker positions clobber the local marker positions. These two position sets will be completely separate.

ZLLentz:
This works essentially flawlessly now, except for on cameras you've never loaded before. I'll round out my day by tracking down why this fails when there isn't a config file to load.

DisplayImage.py
```diff
@@ -466,3 +456,66 @@ def roiSet(self, x, y, w, h, rel=False):
...
+
+def reset_markers(markers: list[param.Point]) -> None:
+    """
+    Given a list of 4 Points that represent marker positions, reset them to the defaults.
+
+    The default marker positions are 100 points horizontally and 100 points
+    vertically away from each of the corners of the image, not on the image
+    itself but off of it.
+    """
+    markers[0].setAbs(-100, 100)
```
janeliu-slac: For `markers[0].setAbs(-100, 100)` should the second value be `-100`?

DisplayImage.py
```diff
@@ -466,3 +456,66 @@ def roiSet(self, x, y, w, h, rel=False):
...
+
+def default_markers() -> list[param.Point]:
+    """
+    Return a list of 4 Points that represent the default marker positions.
+
+    The default marker positions are 100 points horizontally and 100 points
+    vertically away from each of the corners of the image, not on the image
+    itself but off of it.
+    """
+    pts = [param.Point(0, 0), param.Point(0, 0), param.Point(0, 0), param.Point(0, 0)]
```
janeliu-slac, line 505: I'm not too familiar with PyQt. Is it necessary to set pts to `(0, 0)` values before calling `reset_markers(pts)`?

camviewer_ui_impl.py
```diff
@@ -886,41 +877,64 @@ def onSetROI2(self):
...
 
     def onMarkerTextEnter(self, n):
         self.ui.display_image.lMarker[n].setRel(
             float(self.ui.xmark[n].text()), float(self.ui.ymark[n].text())
         )
-        if n <= 1:
-            self.updateMarkerText(False, True, 1 << n, 1 << n)
-        else:
-            self.updateMarkerText(False, True, 0, 1 << n)
+        self.updateMarkerText(False, True, 1 << n, 1 << n)
```
janeliu-slac, line 950: Out of curiosity, why is it shifting the bits left?

DisplayImage.py
ZLLentz: This depends on the coordinate system used and since this app does a lot of stuff internally I have no idea what the actual coordinate system is. I have the same intuition as you that this should be (-100, -100), I tried to copy what was implemented before I got here and didn't think too much about it. I can double-check that resetting the markers puts them in correct-looking positions.

DisplayImage.py
ZLLentz: I think I just wanted to allocate "any" point and the API required an x and a y. `param.Point` isn't actually a qt thing, it's something they wrote into the param.py file in this repo. I could have put in the default values again but I didn't want to implement the defaults in multiple places.

camviewer_ui_impl.py
ZLLentz: Before I made any changes, there was a lot of this annoying bitshift stuff in here and I opted not to refactor it despite it being confusing.

In this case, the `updateMarkerText` method is expecting a bitmask to decide which markers to update, so the caller is bitshifting the number 1 into the right bit slot to only update one marker. There are definitely better ways to do this but I left it as it was prior to the PR.

camviewer_ui_impl.py
```diff
@@ -1865,108 +1868,40 @@ def cross2mon(self, exception=None):
...
-        for i in self.globmarkpvs2:
+        for pv in self.globmarkpvs.values():
             try:
-                i.disconnect()
+                pv.disconnect()
             except Exception:
                 pass
-        self.globmarkpvs2 = []
+        self.globmarkpvs = {}
+        self.ui.cross1_status.setText("L")
```
KaushikMalapati, line 1895: What do these do? Are they supposed to be G when you're in global mode? Ui sets them to L initially, disconnectMarkerPvs sets them to L, but they are never set to G anywhere - presumably they should be in connectMarkerPvs?

KaushikMalapati:
I was trying this on CRIX01 and was swapping between local and global markers in the taskbar when my camviewer became completely uninteractive. When I tried reopening that cam on prod camviewer, it said the pvs were disconnected until I restarted the ioc. I tried this branch on CRIX01 again, got a bunch of disconnect messages and then I had to restart the ioc again.

DisplayImage.py
KaushikMalapati: I also think it should be -100, -100 initially
image

camviewer_ui_impl.py
```diff
@@ -314,10 +305,8 @@ def __init__(
         self.iRangeMax = 1023
         self.camactions = []
         self.lastwidth = 0
-        self.useglobmarks = False
-        self.useglobmarks2 = False
-        self.globmarkpvs = []
-        self.globmarkpvs2 = []
+        self.useglobmarks = True
```
KaushikMalapati, line 308: If you use global markers by default I think the cross#_status widgets in the ui file should be G by default

camviewer_ui_impl.py
```diff
@@ -392,6 +381,8 @@ def __init__(
             None,
         ]
         self.markerdialog.hide()
+        self.local_marker_points = default_markers()
+        self.global_marker_points = default_markers()
```
KaushikMalapati, line 387: What is global_marker_points meant to do? I see it initialized to default_markers and reset in getConfig which make sense, but I don't ever see it set to anything. new_point is set to one of its elements in onCrossUpdate, but won't it always be one of the default points?

camviewer_ui_impl.py
```diff
@@ -1865,108 +1868,40 @@ def cross2mon(self, exception=None):
...
-                i.disconnect()
+                pv.disconnect()
             except Exception:
                 pass
-        self.globmarkpvs2 = []
+        self.globmarkpvs = {}
+        self.ui.cross1_status.setText("L")
+        self.ui.cross2_status.setText("L")
+        self.ui.cross3_status.setText("L")
+        self.ui.cross4_status.setText("L")
```
KaushikMalapati, line 1898: I don't think this is correct because you call disconnectMarkerPvs in connectMarkerPvs to update what they are. This should only set the text to L if useglobmarks

camviewer_ui_impl.py
slactjohnson: They appear to be set to "G" at line 1846 in onCrossUpdate.

ZLLentz: They are set when we get a new value from the cross position PVs. I leave them as "L" until then in case the cross PVs never connect, so that "G" can only be reached after we've definitely connected to a real PV.

camviewer_ui_impl.py
ZLLentz: This is intentional so that they don't turn to G until we successfully connect to and read from the global marker position PVs

camviewer_ui_impl.py
ZLLentz: I swap each global marker point into the live marker points list in display_image when it's time to fill the marker positions with data from the live PVs. This protects local_marker_points from getting filled with global PV values.

camviewer_ui_impl.py
ZLLentz: I think this is always appropriate. I need to guarantee that these say "G" if and only if their value reflects a live PV value, which is never guaranteed until we resolve the PV connection and get the first value.

DisplayImage.py
ZLLentz: I'll fix these, thanks for pointing it out and confirming

ZLLentz:
> swapping between local and global markers in the taskbar when my camviewer became completely uninteractive

This one is concerning to me. I noticed elsewhere that swapping camera really quickly caused freezing behavior, does the same happen when swapping between global and local triggers really quickly?

KaushikMalapati:
> > swapping between local and global markers in the taskbar when my camviewer became completely uninteractive
> 
> This one is concerning to me. I noticed elsewhere that swapping camera really quickly caused freezing behavior, does the same happen when swapping between global and local triggers really quickly?

Yes

ZLLentz:
I'll try to track down which thing I'm doing causes the bad behavior- maybe disconnecting a pyca PV when it is initializing breaks things, for example? It'd be good to document specifically where the deadlock happens.

Probably I'll have to end up (temporarily) disabling certain UI elements after the user changes them.

ZLLentz:
This is how I found (one of) the freezes and identified which line of code we were stuck on:
- start GUI
- select cam via combobox
- mash up and down arrows until it freezes
- py-spy
```
(pcds-5.9.1)zlentz@psbuild-rhel7-01:~$ py-spy dump --pid 6395
Process 6395: python ./camviewer.pyw --instrument xcs --pvlist /cds/group/pcds/pyps/config/xcs/camviewer.cfg
Python v3.9.7 (/cds/group/pcds/pyps/conda/py39/envs/pcds-5.1.1/bin/python3.9)

Thread 6395 (idle): "MainThread"
    disconnect (psp/Pv.py:429)
    disconnectPv (camviewer_ui_impl.py:1784)
    connectCamera (camviewer_ui_impl.py:1911)
    onCameraSelect (camviewer_ui_impl.py:2153)
    <module> (camviewer.pyw:111)
Thread 6418 (idle): "Dummy-16"
```
https://github.com/slaclab/pyca/blob/395124deff292c694ae85e1d040375b6139bfa6e/psp/Pv.py#L429
https://github.com/slaclab/pyca/blob/395124deff292c694ae85e1d040375b6139bfa6e/pyca/pyca.cc#L45

praise `py-spy` for always being there to save me

it looks like the python threads are just chilling here waiting for pyca's `clear_channel` to complete
maybe it's still not obvious how to avoid this other than forcing the user to slow down

there could be "something" to check when we disconnect Pv objects to know to wait a moment

DisplayImage.py
ZLLentz: ok yes I also see now that I made a copy/paste error somehow when putting this all in one place, whoops!

ZLLentz:
For continuity, here's the py-spy dump from the global marks spam method, which I couldn't get to happen on psbuild but definitely happens for CRIX01 from rix-control
```
$ py-spy dump --pid 11118
Process 11118: python ./camviewer.pyw --instrument rix --pvlist /cds/group/pcds/pyps/config/rix/camviewer.cfg
Python v3.9.7 (/cds/group/pcds/pyps/conda/py39/envs/pcds-5.1.1/bin/python3.9)

Thread 11118 (idle): "MainThread"
    disconnect (psp/Pv.py:429)
    disconnectMarkerPVs (camviewer_ui_impl.py:1896)
    setUseGlobalMarkers (camviewer_ui_impl.py:928)
    onGlobMarks (camviewer_ui_impl.py:906)
    local_clicked (camviewer_ui_impl.py:888)
    <module> (camviewer.pyw:111)
Thread 11434 (idle): "Dummy-8"
```
Very similar

In offline testing, the only thing I've been able to identify is that calling `disconnect` on a `psp.Pv` object with a monitor callback running waits until the monitor callback finishes. Possibly this is the source of the deadlock for specific monitor callbacks.

ZLLentz:
One of my other windows froze on close at a different part:
```
$ py-spy dump --pid 8437
Process 8437: python ./camviewer.pyw --instrument rix --pvlist /cds/group/pcds/pyps/config/rix/camviewer.cfg
Python v3.9.7 (/cds/group/pcds/pyps/conda/py39/envs/pcds-5.1.1/bin/python3.9)

Thread 8437 (idle): "MainThread"
    unsubscribe (psp/Pv.py:506)
    monitor_stop (psp/Pv.py:883)
    end_monitors (camviewer_ui_impl.py:754)
    closeEvent (camviewer_ui_impl.py:726)
    <module> (camviewer.pyw:111)
Thread 8480 (idle): "Dummy-2"
```

ZLLentz:
I was not super happy with any of the options I tried, so I settled on one that's at least simple:

- When the user does an action that might freeze the GUI if done multiple times quickly, disable the associated widgets for 1 second

In the process of getting to this point I tidied up much of the PV setup and cleanup code here, which all are probably good improvements but were red herrings as far as solving the base issue here.

There were multiple flavors of GUI freezing (e.g. getting stuck on a qt signal emit, getting stuck on literally nothing (?), getting stuck on app cleanup) and I was unable to figure out how to wrangle all of (any of?) them.

camviewer_ui_impl.py
```diff
@@ -1200,35 +1213,20 @@ def onComboBoxColorIndexChanged(self, iNewIndex):
...
-            except Exception:
-                pass
-            self.lensPv = None
-        if self.putlensPv is not None:
-            try:
-                self.putlensPv.disconnect()
-            except Exception:
-                pass
-            self.putlensPv = None
+        self.camera = self.disconnectPv(self.camera)
```
tangkong, line 1216: To clarify how this works, we: 
* set these attributes to instances of a `psp.Pv.Pv` on startup (self.connectCamera)
* do normal things
* set these to `None` via `self.disconnectPv` as a way of deallocating the resources.

I thought it was a bit opaque, but didn't take too long to backtrack.  I think maybe a stray comment might have helped, but I think it's not a huge deal

camviewer_ui_impl.py
```diff
@@ -2415,6 +2336,9 @@ def buttonWriteCallback(self, button):
 
     def setupSpecific(self):
         self.ui.actionGlobalMarkers.setChecked(self.useglobmarks)
+        self.ui.global_button.setChecked(self.useglobmarks)
+        self.ui.local_button.setChecked(not self.useglobmarks)
```
tangkong, line 2340: Nit/Rabbithole: Apparently the way to give multiple radio buttons exclusive behaviour is to group them in a [`QButtonGroup`](https://doc.qt.io/qtforpython-5/PySide2/QtWidgets/QButtonGroup.html)

I could have sworn there was an easier way to do this but I guess not.

camviewer_ui_impl.py
```diff
@@ -3067,40 +2956,70 @@ def getConfig(self):
...
-        newheight = self.cfg.viewheight
-        if int(newwidth) < self.minwidth:
-            newwidth = str(self.minwidth)
-        if int(newheight) < self.minheight:
-            newheight = str(self.minheight)
-        newproj = self.cfg.projsize
-        self.advdialog.ui.viewWidth.setText(newwidth)
-        self.advdialog.ui.viewHeight.setText(newheight)
-        self.advdialog.ui.projSize.setText(newproj)
+        # viewwidth may be missing if there was no camera config
```
tangkong, line 2959: There are a lot of instances here of catching missing configs.  Perhaps the proper way to handle this is to bundle a default config file and refer to it whenever an external one is not provided.  That way we can be certain that a value exists and we can manage those values easily (rather than setting them in code?)

That being said we were supposed to stop doing camviewer work ages ago.

camviewer_ui_impl.py
```diff
@@ -3119,90 +3041,113 @@ def getConfig(self):
...
-            self.set_new_min_pixel(int(self.cfg.colormin))
-            self.set_new_max_pixel(int(self.cfg.colormax))
+            colorscale = self.cfg.colorscale
         except Exception:
-            print("Failed to load min or max pixel value from config file.")
+            # Same default as the combobox in the UI file
+            colorscale = ["Linear", "Scale"]
+        finally:
+            # Backcompat for really really really old names
+            if colorscale[0] == "Log":
```
tangkong, line 3068: What do people have against `Ln` for natrual log? 🤔

camviewer_ui_impl.py
```diff
@@ -886,41 +879,75 @@ def onSetROI2(self):
...
+
+    def allow_glob_changes(self, allow: bool = True):
+        """
+        Helper function to briefly disable all the global vs local widgets.
+        """
+        self.ui.actionGlobalMarkers.setEnabled(allow)
+        self.ui.global_button.setEnabled(allow)
+        self.ui.local_button.setEnabled(allow)
+        if not allow:
+            self.glob_changeable_restore.singleShot(1000, self.allow_glob_changes)
```
tangkong, line 918: I supposed the alternative is to do some lock acquisition and disabling of the UI whenever a lengthy action is initiated?  That's a bit more involved of a refactor though.  

Stray thought: a busy cursor would probably be appreciated here if we start to deal with many lengthy actions

tangkong (approved):
Code looks good to me on the whole.  I think I only have nitpicks and shower thoughts to contribute.  

I took it for a spin and couldn't get it to crash/lock up in the old or new ways

camviewer_ui_impl.py
ZLLentz: You've got it right. To be clear, I'm not a big fan of this pattern that's repeated dozens of times up and down the application but I also wasn't interested in doing the work to replace it with something else.

camviewer_ui_impl.py
ZLLentz: The qt docs said that I could just put them in the same layout but the exclusivity didn't seem to work specifically in the case where the user changed global mode from the top menu.

camviewer_ui_impl.py
ZLLentz: Yeah, that's a lot cleaner. I'll keep this idea in mind in general for config-reading applications in the future, if not for this one.

camviewer_ui_impl.py
ZLLentz: There aren't any lengthy actions here, rather there's an annoying race condition that comes from pyca:

- `disconnect` and `unsubscribe` both block and wait for any ongoing subscription callbacks to complete
- Something in the related cleanup causes the subscription callback to block
- Both threads block forever

I think it's a bilateral deadlock. It seems to have a few flavors, sometimes a signal.emit call hangs forever, for example.

tangkong: Hm, troubling.  I can only assume there's some more involved way to unravel the chain of execution and ensure everything waits for its predecessors, but that's a task for later

ZLLentz: Yeah I couldn't figure it out after trying for an hour or two, so I went for this plan B "Avoid the problem"

camviewer_ui_impl.py
ZLLentz: To clarify: clicking the button reflected exclusivity, but updating the checked state myself did not.

PR #35, opened 2025-02-06
ENH: remove nord and nelm and adjust accordingly

ZLLentz:
### Main Goal
- Remove .NORD and .NELM references as they are redundant and seem to get dropped by our gateway servers.
- Use the array size information from area detector PVs to set the array counts wherever we were doing this previously

### Changes to `__init__`
This is called at GUI startup.

- remove references to now unused norm, nelm pvs and max counts.

### Changes to `wantImage`
This is called whenever it's time to collect and render a new image.

- Replace nord -> count pipeline with a row * col -> count pipeline (or *3 if color cam)
- Pass count into the PV acquisition in case count has changed (unreported bug, fixed)

### Changes to `connectCamera`
This is called whenever someone selects a new camera to initialize, if the camera looks like its online.

- Move EPICS-sourced-string re-inits to empty string to `self.clear`, which where we moved the `disconnectPv` calls in #34 
- Remove nord, nelm pvs
- Clean up the sizing information initialization
- Fix an issue where the wrong bit count PV was used and clean up the bit count setup
- Remove vestigial self.scale
- General cleanup

And last and maybe most important, I rearranged the initilaization order:

#### Old order
1. Nord/nelm -> count
2. Connect to image PV
3. Get xy, color, bit depth info

#### New order
1. Connect to `x*y*color` -> count
2. Connect to bit depth PV, get bit depth
3. Connect to image PV

### DIFF
--- a/camviewer_ui_impl.py
+++ b/camviewer_ui_impl.py
@@ -2032,52 +2032,71 @@ class GraphicUserInterface(QMainWindow):
         size0 = self.connectPv(self.cameraBase + ":ArraySize0_RBV")
         size1 = self.connectPv(self.cameraBase + ":ArraySize1_RBV")
         size2 = self.connectPv(self.cameraBase + ":ArraySize2_RBV")
-        if None in (size0, size1):
-            # Note: allow B/W cams to not have a size2 PV
-            self.ui.label_status.setText("IOC timeout in setup")
-            print("IOC timeout in setup (image size PVs)")
-            return
 
-        if size2 is None or size2.value == 0:
+        def get_size_val(pv_or_none: Pv | None) -> int | None:
+            """
+            Get the actual size given the Pv after connectPv.
+
+            Disconnected PVs will return None
+            Bad values will return 0
+            """
+            if pv_or_none is None:
+                return None
+            try:
+                val = int(pv_or_none.value)
+            except Exception:
+                val = 0
+            return max(0, val)
+
+        size0_val = get_size_val(size0)
+        size1_val = get_size_val(size1)
+        size2_val = get_size_val(size2)
+
+        def show_sizing_error(error_text: str):
+            self.disconnectPv(size0)
+            self.disconnectPv(size1)
+            self.disconnectPv(size2)
+            # Clear the image so that we don't have a stale image from the previous cam
+            self.ui.display_image.image.fill(0)
+            self.ui.display_image.repaint()
+            # Report which issue we had
+            self.ui.label_status.setText(error_text)
+            print(f"{error_text} (rows/cols)")
+
+        if None in (size0_val, size1_val):
+            # pvs must be connected
+            return show_sizing_error("IOC timeout")
+        elif 0 in (size0_val, size1_val):
+            # pvs must be nonzero
+            return show_sizing_error("Zero pixels in image")
+        elif size0_val == 3 and size2_val is None:
+            # Ambiguous: either disconnected pv in color cam
+            # or really strange IOC config with missing pv
+            return show_sizing_error("Ambiguous sizing")
+        elif size0_val != 3 and size2_val is not None and size2_val > 0:
+            # Weird multidimensional thing??
+            return show_sizing_error("Invalid cam dimensions")
+
+        # No errors, proceed!
+        if size2_val in (0, None):
             # Just B/W!
             self.rowPv = size1
             self.colPv = size0
             self.disconnectPv(size2)
             self.isColor = False
-        elif size0.value == 3:
+        elif size0_val == 3:
             # It's a color camera!
             self.rowPv = size2
             self.colPv = size1
             self.disconnectPv(size0)
             self.isColor = True
         else:
-            # Something else? e.g. 100x100x100
-            self.ui.label_status.setText("Invalid cam dimemsions")
-            print("Invalid cam dimensions in setup (3d or more)")
-            return
+            # It shouldn't be possible to get here, but if we do...
+            return show_sizing_error("Unknown sizing error")
 
-        # See if we've connected to a camera with valid height and width
-        if (
-            self.rowPv is None
-            or self.rowPv.value == 0
-            or self.colPv is None
-            or self.colPv.value == 0
-        ):
-            # Clear the image so that we don't have a stale image from the previous cam
-            self.ui.display_image.image.fill(0)
-            self.ui.display_image.repaint()
-            # Report which issue we had
-            if self.rowPv is None or self.colPv is None:
-                self.ui.label_status.setText("IOC timeout in setup")
-                print("IOC timeout in setup (rows/cols)")
-            else:
-                self.ui.label_status.setText("Zero pixels in image")
-                print("Zero pixels in image (no width/height)")
-            return
-        else:
-            self.count = self.rowPv.value * self.colPv.value
-            if self.isColor:
-                self.count *= 3
+        self.count = self.rowPv.value * self.colPv.value
+        if self.isColor:
+            self.count *= 3
 
         if self.lFlags[index] != "":
             self.bits = int(self.lFlags[index])
--- a/camviewer_ui_impl.py
+++ b/camviewer_ui_impl.py
@@ -276,7 +276,7 @@ class GraphicUserInterface(QMainWindow):
         # Default to VGA!
         param.setImageSize(640, 480)
         self.isColor = False
-        self.bits = 10
+        self.bits = 12
         self.maxcolor = 1023
         self.lastUpdateTime = time.time()
         self.dispUpdates = 0
@@ -2083,11 +2083,15 @@ class GraphicUserInterface(QMainWindow):
             self.bits = int(self.lFlags[index])
         else:
             self.bits_pv = self.connectPv(self.cameraBase + ":BitsPerPixel_RBV")
-            if self.bits_pv is None:
-                self.ui.label_status.setText("IOC timeout in setup")
-                print("IOC timeout in setup (bits per pixel PV)")
-                return
-            self.bits = self.bits_pv.value
+            try:
+                self.bits = int(self.bits_pv.value)
+            except Exception:
+                self.bits = 12
+                print("Bits PV did not connect or had bad value, using default 12")
+
+        # Ensure positive bit depth no bigger than 32
+        # Negative bit depth and very large bit depths both break the app
+        self.bits = min(max(1, self.bits), 32)
         self.maxcolor = (1 << self.bits) - 1
         self.ui.horizontalSliderRangeMin.setMaximum(self.maxcolor)
         self.ui.horizontalSliderRangeMin.setTickInterval((1 << self.bits) / 4)
--- a/camviewer_ui_impl.py
+++ b/camviewer_ui_impl.py
@@ -1953,49 +1953,25 @@ class GraphicUserInterface(QMainWindow):
         size1 = self.connectPv(self.cameraBase + ":ArraySize1_RBV")
         size2 = self.connectPv(self.cameraBase + ":ArraySize2_RBV")
 
-        def get_size_val(pv_or_none: Pv | None) -> int | None:
-            """
-            Get the actual size given the Pv after connectPv.
+        size0_val = self._get_size_val(size0)
+        size1_val = self._get_size_val(size1)
+        size2_val = self._get_size_val(size2)
 
-            Disconnected PVs will return None
-            Bad values will return 0
-            """
-            if pv_or_none is None:
-                return None
-            try:
-                val = int(pv_or_none.value)
-            except Exception:
-                val = 0
-            return max(0, val)
-
-        size0_val = get_size_val(size0)
-        size1_val = get_size_val(size1)
-        size2_val = get_size_val(size2)
-
-        def show_sizing_error(error_text: str):
-            self.disconnectPv(size0)
-            self.disconnectPv(size1)
-            self.disconnectPv(size2)
-            # Clear the image so that we don't have a stale image from the previous cam
-            self.ui.display_image.image.fill(0)
-            self.ui.display_image.repaint()
-            # Report which issue we had
-            self.ui.label_status.setText(error_text)
-            print(f"{error_text} (rows/cols)")
+        error_info = (size0, size1, size2, size0_val, size1_val, size2_val)
 
         if None in (size0_val, size1_val):
             # pvs must be connected
-            return show_sizing_error("IOC timeout")
+            return self._show_sizing_error("IOC timeout", error_info)
         elif 0 in (size0_val, size1_val):
             # pvs must be nonzero
-            return show_sizing_error("Zero pixels in image")
+            return self._show_sizing_error("Zero pixels in image", error_info)
         elif size0_val == 3 and size2_val is None:
             # Ambiguous: either disconnected pv in color cam
             # or really strange IOC config with missing pv
-            return show_sizing_error("Ambiguous sizing")
+            return self._show_sizing_error("Ambiguous sizing", error_info)
         elif size0_val != 3 and size2_val is not None and size2_val > 0:
             # Weird multidimensional thing??
-            return show_sizing_error("Invalid cam dimensions")
+            return self._show_sizing_error("Invalid cam dimensions", error_info)
 
         # No errors, proceed!
         if size2_val in (0, None):
@@ -2012,7 +1988,7 @@ class GraphicUserInterface(QMainWindow):
             self.isColor = True
         else:
             # It shouldn't be possible to get here, but if we do...
-            return show_sizing_error("Unknown sizing error")
+            return self._show_sizing_error("Unknown sizing error", error_info)
 
         self.count = self.rowPv.value * self.colPv.value
         if self.isColor:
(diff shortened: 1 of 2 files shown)


ZLLentz:
Writing out the PR description here clued me in some other small adjustments to make prior to review

KaushikMalapati:
We are getting rid of .NORD and .NELM pvs in part because of gateway issues, but are also adding news :ArraySize0_RBV and :BitsPerPixel_RBV pvs. Can they have the same problem since they are also not going through the cam gateways?

camviewer_ui_impl.py
```diff
@@ -2098,6 +2060,34 @@ def connectCamera(self, sCameraPv, index, sNotifyPv=None):
...
+        self.maxcolor = (1 << self.bits) - 1
+        self.ui.horizontalSliderRangeMin.setMaximum(self.maxcolor)
+        self.ui.horizontalSliderRangeMin.setTickInterval((1 << self.bits) / 4)
+        self.ui.horizontalSliderRangeMax.setMaximum(self.maxcolor)
+        self.ui.horizontalSliderRangeMax.setTickInterval((1 << self.bits) / 4)
+        self.ui.spinbox_range_max.setMaximum(self.maxcolor)
+        self.ui.spinbox_range_min.setMaximum(self.maxcolor)
+
+        # Try to connect to the camera
+        self.camera = self.connectPv(sCameraPv, count=self.count)
```
KaushikMalapati, line 2023: What is the reason to provide count here if we recalculate count in wantImage? I assume the recalculation there is in case it ever changes.

camviewer_ui_impl.py
```diff
@@ -1424,15 +1422,10 @@ def wantImage(self, want=True):
...
-                if self.nordPv:
-                    self.count = int(self.nordPv.value)
-                    if self.count == 0:
-                        sz = int(self.rowPv.value) * int(self.colPv.value)
-                        if sz > 0 and sz < self.maxcount:
-                            self.count = sz
-                        else:
-                            self.count = self.maxcount
-                self.camera.get(timeout=None)
+                self.count = self.rowPv.value * self.colPv.value
```
KaushikMalapati: Does it make sense to make this dependent on rowPv and colPv being connected? Don't we always want all the entire waveform record?

camviewer_ui_impl.py
```diff
@@ -2027,58 +2019,28 @@ def connectCamera(self, sCameraPv, index, sNotifyPv=None):
...
-            self.maxcount = int(self.nelmPv.value)
-        except Exception:
-            self.nelmPv = None
-            self.maxcount = None
-        if self.count is None or self.count == 0:
-            self.count = self.maxcount
-        self.camera = self.connectPv(sCameraPv, count=self.count)
-        if self.camera is None:
+        # Try to get the camera size!
+        self.scale = 1
```
KaushikMalapati: Is self.scale ever not one?

camviewer_ui_impl.py
```diff
@@ -2027,58 +2019,28 @@ def connectCamera(self, sCameraPv, index, sNotifyPv=None):
...
+        size0 = self.connectPv(self.cameraBase + ":ArraySize0_RBV")
+        size1 = self.connectPv(self.cameraBase + ":ArraySize1_RBV")
+        size2 = self.connectPv(self.cameraBase + ":ArraySize2_RBV")
+        if None in (size0, size1, size2):
```
KaushikMalapati: Is it possible we have a B/W camera that could word without :ArraySize2_RBV or just doesn't have it but this causes connectCamera to unnecessarily fail?

camviewer_ui_impl.py
```diff
@@ -2098,6 +2060,34 @@ def connectCamera(self, sCameraPv, index, sNotifyPv=None):
...
             return
+        else:
+            self.count = self.rowPv.value * self.colPv.value
+            if self.isColor:
+                self.count *= 3
+
+        if self.lFlags[index] != "":
+            self.bits = int(self.lFlags[index])
+        else:
+            self.bits_pv = self.connectPv(self.cameraBase + ":BitsPerPixel_RBV")
```
KaushikMalapati, line 2000: self.bits is set to 10 in init. What would be the negative consequences of using ten if we don't get the real number of bits from the camviewer.cfg or bits_pv instead of just failing to connect?

KaushikMalapati:
Did you ever try this? https://jira.slac.stanford.edu/browse/ECS-7167

KaushikMalapati:
Overall good. I have some nitpicks that I left comments about, mostly concerning things that would cause a camera to not connect or the camviewer not to update.

camviewer_ui_impl.py
```diff
@@ -2098,6 +2060,34 @@ def connectCamera(self, sCameraPv, index, sNotifyPv=None):
...
+            if self.bits_pv is None:
+                self.ui.label_status.setText("IOC timeout in setup")
+                print("IOC timeout in setup (bits per pixel PV)")
+                return
+            self.bits = self.bits_pv.value
```
KaushikMalapati: I don't think it should ever not be, but is this value guaranteed to be positive?

ZLLentz:
> We are getting rid of .NORD and .NELM pvs in part because of gateway issues, but are also adding news :ArraySize0_RBV and :BitsPerPixel_RBV pvs. Can they have the same problem since they are also not going through the cam gateways?

I've so far only observed obscure gateway issues (stale values) for the dot-led fields attached to the ArrayData record, I haven't ever seen this for completely separate records, though I have seen a spattering of fully disconnected PVs through the gateway in the middle of a camera IOC, so it's good to be skeptical of adding PVs in.

It's worth nothing that these PVs aren't exactly new, I'm just using them differently. I think the diff is pretty clear on the old use of ArraySize0, but BitsPerPixel was particularly broken on the previous version, instead sometimes deferring to a no-longer-used BIT_DEPTH PV (which doesn't show up in grep_pv at all)

camviewer_ui_impl.py
ZLLentz: If we pass count here, the camera array PV will collect exactly that many data points (up to what's actually available in the buffer). If we don't pass count here, it will use the default value for count, which in this case is "whatever count was when we set up `self.camera`". I added this in because I was worried about our array size getting out of sync with the camera size if the image size changes at runtime (e.g. due to ROI or binning changes)

Note that I found I still needed to pass `count` to the initialization of `self.camera` because it affected the initial sampling of the array for some of the things we're doing on camera setup.

camviewer_ui_impl.py
ZLLentz: You're right, I should be more careful here. I think `value` is guaranteed to be populated for each of these if we reach this stage because I stop setting up the camera if these don't get good values at camera select. Therefore if they disconnect later we still have the last known good value, but there could be other unexpected issues from trusting that they exist and are well-formed. I'll think a little on what kind of useful guard rails to add.

The waveform record contains NELM elements, of which NORD of them are valid data points. If you simply get the array without passing count, `pyca` should respect NORD (I think?). Since I can't trust NORD and NELM, I select the count manually using the other sizing information.

camviewer_ui_impl.py
ZLLentz: This is a vestigial part of the app that I didn't clean up. I'll clean it up now since I think it doesn't get used in any meaningful way right now (there's code in a few places that divides numbers by `self.scale`)

camviewer_ui_impl.py
ZLLentz: I was worried about this too, so I ran a one-line bash loop:
```
for num in {0..9}; do grep_pv "ArraySize${num}_RBV" | wc -l; done
```
And found that for currently deployed systems:
- most cameras have all values from 0 to 9 (but they leave the unused dimensions unpopulated)
- some cameras have 0, 1, and 2 but not the rest
- no cameras have only 0 and 1

But I think your skepticism is warranted, after all, what if a camera behaves like this in the future?

We should be allowed to proceed if size2 is missing so long as size0 isn't 3 (so it's a B/W cam). I'll revise to guard against that case.

camviewer_ui_impl.py
ZLLentz: This is pretty low stakes I actually, I thought it was used in more places but this is only used to set the bounds for the color mapping. The image processing internals always allocate 32 bits regardless.

If `self.bits` is too big, we'll have extra space in the color map sliders, which isn't so bad.
If `self.bits` is too small, we won't be able to see the full range of the bit depth in the image.

I think it's a judgement call whether or not to allow us to proceed without this PV. A missing PV here is likely an IOC or configuration issue, but the consequence is pretty minor. Maybe I should increase the default from 10 to 12 to match some of these newer gige models and allow it to fail and fall back to 12.

ZLLentz: One more related issue, the camviewer as written doesn't ever update this. So, if you reconfigure your camera to have a higher bit depth while viewing the camera, you can't see those in camviewer unless you re-initialize the PV connections...

camviewer_ui_impl.py
ZLLentz: I think it's prudent to set some bounds of this just in case.
Maybe 8 as a minimum, 32 as a maximum?

ZLLentz:
Thank you for your review Kaushik, that was all very helpful and I'll be making adjustments

ZLLentz:
> https://jira.slac.stanford.edu/browse/ECS-7167

I haven't tried this yet, I'd prefer to wait for the next PAMM

camviewer_ui_impl.py
KaushikMalapati: I wonder if you can make the roi small enough that a black and white camera only has three columns

ZLLentz: I think you actually can, in which case we'd need to check if ArraySize2 is nonzero instead as the color cam check. That might be worth doing.

camviewer_ui_impl.py
KaushikMalapati: I would've just done > 0, but 8-32 is probably more reasonable

camviewer_ui_impl.py
ZLLentz: Some example sizes for reference:
B/W
```
$ caget IM3L0:PPM:CAM:IMAGE2:ArraySize{0..2}_RBV
IM3L0:PPM:CAM:IMAGE2:ArraySize0_RBV 512
IM3L0:PPM:CAM:IMAGE2:ArraySize1_RBV 512
IM3L0:PPM:CAM:IMAGE2:ArraySize2_RBV 0
```
Color
```
caget XCS:GIGE:08:IMAGE2:ArraySize{0..2}_RBV
XCS:GIGE:08:IMAGE2:ArraySize0_RBV 3
XCS:GIGE:08:IMAGE2:ArraySize1_RBV 1292
XCS:GIGE:08:IMAGE2:ArraySize2_RBV 964
```

camviewer_ui_impl.py
KaushikMalapati: Maybe set it to 12 as a default, and let it fail with a console message? That way a user would know about the problem without restricting their camviewer usage completely. 

Updating the record should probably be an issue on ioc-common-gigecam. I guess we would also need a callback to update the range limits if the pv ever updates too, but that's in the future if ever.

camviewer_ui_impl.py
ZLLentz: I picked 8 because an image with less than 256 values per pixel seems wild
I picked 32 because that's what the viewer allocates internally, so above that it won't work properly anyway.


I mostly want an upper bound to avoid situations like:
- bit depth is erroneously 2 billion
- color range max gets set to 2^2billion

ZLLentz: let's let it be bit depth 1-32 just in case someone has a really strange camera one day

camviewer_ui_impl.py
ZLLentz: I think the gigecam IOC already allows you to live update the bit depth, so it's mostly on this side we could add a callback in the future

ZLLentz:
I've updated in response to all of Kaushik's comments, but I'm seeing some odd behavior on color cams that I need to resolve.

Examples:
- xcs_gige_8 crashes the gui if you increase the color range max to the full maximum
  - reports 24 bits per pixel, has pixels in the range 14-630
- xcs_gige_22 when viewed in greyscale mode is limited to 8 bits by camviewer but it actually has more
  - reports 8 bits per pixel, has pixels in the range  340-420

It's interesting to note that in the previous version color cams were always (accidentally?) forced into the default bit depth of 10

ZLLentz:
The crash comes because internally the camviewer only allocates enough memory for a 16-bit colormap, despite allocating for 32-bit values elsewhere. I'll keep this PR simple and reduce the max bit depth from 32 to 16.

I suspect that 8 bits per pixel in color goes from [0, 255] to [0, 255*3] = [0, 765] because we just sum the color values. This has implications for the correct setting of the maximum color map value. I'll adjust accordingly and see how things are afterwards.

This also implies that that there are other internal issues here but I'm going to gloss over that for now.

ZLLentz:
I think this is better now

camviewer_ui_impl.py
```diff
@@ -2089,14 +2089,18 @@ def connectCamera(self, sCameraPv, index, sNotifyPv=None):
...
+        # Negative bit depth and large bit depths both break the app
+        self.bits = min(max(1, self.bits), 16)
+        self.maxcolor = 2**self.bits - 1
+        if self.isColor:
+            self.maxcolor *= 3
+            self.maxcolor = min(self.maxcolor, 2**16 - 1)
+
         self.ui.horizontalSliderRangeMin.setMaximum(self.maxcolor)
-        self.ui.horizontalSliderRangeMin.setTickInterval((1 << self.bits) / 4)
+        self.ui.horizontalSliderRangeMin.setTickInterval(self.maxcolor // 4)
```
KaushikMalapati, line 2016: Do we need explicit floor division? Nothing wrong with it, just curious

camviewer_ui_impl.py
```diff
@@ -2032,23 +2032,29 @@ def connectCamera(self, sCameraPv, index, sNotifyPv=None):
...
             print("IOC timeout in setup (image size PVs)")
             return
 
-        if size0.value == 3:
+        if size2 is None or size2.value == 0:
+            # Just B/W!
+            self.rowPv = size1
+            self.colPv = size0
+            self.disconnectPv(size2)
+            self.isColor = False
```
KaushikMalapati: I think we could hit a case where we have a color cam with size0 = 3 and size2 disconnected, and this block would think it's a black and white camera. 

Can we just have the check here be `size2.value == 0` which I think still allows black and white cameras to let size2 be disconnected, and then add an elif block affter but before the `elif size0.value == 3` that checks for size2 being disconnected and with the same body as the None check for size0 and size1?

camviewer_ui_impl.py
ZLLentz: In python3.9 this can be a float, but it must be an int for python3.10 thanks to https://github.com/slaclab/pydm/issues/951

Actually I guess I didn't check that floor div gave an int, maybe I should do that (edit: it gives the same type as your inputs)
Either way it doesn't matter much until we update versions, just a habit of mine after that fiasco

camviewer_ui_impl.py
ZLLentz: Your right, there would be a race condition here in e.g. some bad networking cases that could be confusing.

I think the details of the fix are fuzzier. `size2.value` is an `AttributeError` if `size2` is disconnected, for example.

ZLLentz: I think what I want to do is carve out a new error case for 3 x N x disconnected and leave the rest of the logic similar. So overall we have error cases for:
- size0 or size1 being disconnected or 0
- size0 = 3 with size2 disconnected
- size0 !=3 with size2 > 0

ZLLentz: I think what I'm actually ending up writing here has elements of your comment too

ZLLentz:
I think reorganizing with these helper functions and checking for all the error states first made the code easier to read.

ZLLentz:
I'm no longer interested in releasing this on a Friday, so I'm hoping to look again with fresh eyes on Monday etc.

camviewer_ui_impl.py
```diff
@@ -2032,52 +2032,71 @@ def connectCamera(self, sCameraPv, index, sNotifyPv=None):
...
+            # Report which issue we had
+            self.ui.label_status.setText(error_text)
+            print(f"{error_text} (rows/cols)")
+
+        if None in (size0_val, size1_val):
+            # pvs must be connected
+            return show_sizing_error("IOC timeout")
+        elif 0 in (size0_val, size1_val):
+            # pvs must be nonzero
+            return show_sizing_error("Zero pixels in image")
```
KaushikMalapati: If only size0_val is zero and it's a color cam the error text would be slightly misleading

KaushikMalapati:
I think this is good, but I'll take another look Monday and maybe we could wait another day or two in case the other reviewers catch something I missed.

ZLLentz:
I'll add Robert as a reviewer next week too, he's supposed to be back from vacation on Monday

camviewer_ui_impl.py
```diff
@@ -2027,76 +2028,106 @@ def connectCamera(self, sCameraPv, index, sNotifyPv=None):
...
-            return
-
         # Try to get the camera size!
-        self.scale = 1
-        if caget(self.cameraBase + ":ArraySize0_RBV") == 3:
+        size0 = self.connectPv(self.cameraBase + ":ArraySize0_RBV")
+        size1 = self.connectPv(self.cameraBase + ":ArraySize1_RBV")
+        size2 = self.connectPv(self.cameraBase + ":ArraySize2_RBV")
+
+        def get_size_val(pv_or_none: Pv | None) -> int | None:
```
tangkong: Since this is scoped inside the `connectCamera` method, won't it be re-defined every time the method is called?  

Was there a reason to not just make this a standalone helper / method?

camviewer_ui_impl.py
```diff
@@ -2027,76 +2028,106 @@ def connectCamera(self, sCameraPv, index, sNotifyPv=None):
...
+                val = int(pv_or_none.value)
+            except Exception:
+                val = 0
+            return max(0, val)
+
+        size0_val = get_size_val(size0)
+        size1_val = get_size_val(size1)
+        size2_val = get_size_val(size2)
+
+        def show_sizing_error(error_text: str):
```
tangkong: Same thought here as in `get_size_val`, it seems strange to me to have this be a locally scoped function

camviewer_ui_impl.py
```diff
@@ -2027,76 +2028,106 @@ def connectCamera(self, sCameraPv, index, sNotifyPv=None):
...
+            self.disconnectPv(size0)
+            self.disconnectPv(size1)
+            self.disconnectPv(size2)
+            # Clear the image so that we don't have a stale image from the previous cam
+            self.ui.display_image.image.fill(0)
+            self.ui.display_image.repaint()
+            # Report which issue we had
+            self.ui.label_status.setText(error_text)
+            print(f"{error_text} (rows/cols)")
```
tangkong: Currently all these error messages have no row/col information.  Should we be grabbing that info in this function or feeding it in through the error text?

ZLLentz:
I need to do some merge conflict resolution but hopefully it ends up not being too bad

ZLLentz:
It was more than just a trivial merge so I'm going to test again interactively

tangkong (commented):
Dropping these comments off before I get sidetracked again.  

Kaushik did a great job reviewing this, better than I could have I bet.  I'll try to do some interactive testing / monkey keyboard slamming before I approve

camviewer_ui_impl.py
ZLLentz: Yeah I'm absolutely just being lazy with both of these. I'll make them standalone.

camviewer_ui_impl.py
ZLLentz: I could add more info, right now it's just letting you know that this issue happened in the rows/cols step (for example, if a rows/col PV was disconnected vs some other PV)

ZLLentz:
I think all comments are addressed now, and I tested again again

KaushikMalapati (approved):

camviewer_ui_impl.py
ZLLentz: You're right, but I'm going to hope this doesn't happen because I want to work on other tickets. I think it should be pretty rare, if it happens at all.

tangkong (approved):
LGTM, may camviewer find rest

PR #36, opened 2025-02-06
FIX: if initial selected camera is not ready, wait up to 3s

ZLLentz:
As it says in the title.

I chose not to use the psp wait primitives because they let us through prior to the connection callbacks being run. I then opted for a less sophisticated but more simple wait loop.

I've verified that several live cameras connect successfully and that the GUI waits a little extra before giving up for cameras that are truly disconnected (on startup)


ZLLentz:
More reviewers than normal in hopes of a quick merge rather than in hope of a deep review

silkenelson (approved):
This looks good to me, but I have only looked at the code and not performed tests on the real system. Thank you for the quick turn around!

ZLLentz:
I'll merge, roll out a bugfix tag, and if it isn't sufficient we can go again.

PR #38, opened 2025-10-07
FIX: allow camviewer to process color images with large pixels

ZLLentz:
Ref https://jira.slac.stanford.edu/browse/ECS-8777

camViewer was failing to render images from cameras using vimbaCam
After some debug I noticed that the pixel values in the resulting image were strange, with zero values in every other slot, e.g.:
rgb = (5, 0, 5)
rgb = (0, 5, 0)
rgb = (6, 0, 4)

I tracked this down to the image data type being SHORT (2 bytes per element) instead of CHAR (1 byte per element)

I duplicated the logic used to handle these higher-sized arrays for mono cameras into the color cams section

Before:

vimba_color_broken

After:

vimba_color_working

Possibly we could do some more testing but I wanted to put this up before stopping for the day

### DIFF
--- a/pycaqtimage/pycaqtimage.sip
+++ b/pycaqtimage/pycaqtimage.sip
@@ -396,10 +396,10 @@ void _pyDoAvg(ImageBuffer *imageBuffer, T *cadata)
 	_pyCopyToQImage(imageBuffer, 1);
 }
 
-static void _pyColorImagePvCallback(void* cadata, long count, size_t size, void* usr)
+template <class T>
+void _pyDoAvgColor(ImageBuffer *imageBuffer, T *cadata)
 {
-    ImageBuffer*  imageBuffer = reinterpret_cast<ImageBuffer*>(usr);
-    uint8_t*  src    = reinterpret_cast<uint8_t*>(cadata); /* R, G, B! */
+    T*  src    = cadata; /* R, G, B! */
     uint32_t *dst    = imageBuffer->imageData;
     float    *dstF   = imageBuffer->imageDataF;
     int orientation  = imageBuffer->orientation;
@@ -412,13 +412,6 @@ static void _pyColorImagePvCallback(void* cadata, long count, size_t size, void*
     int col_inc      = imageBuffer->srcwidth * colIncMult[orientation] +
                        colIncK[orientation];
     int iNewAverage  = imageBuffer->iNumAveraged + 1;
-
-    UNUSED(size);
-
-    if (count != imageBuffer->size * 3) {
-        // Wrong data size, unsafe to continue
-        return;
-    }
 
     /* If we're using the color image, don't average, just copy and we're done! */
     if (!imageBuffer->useGray) {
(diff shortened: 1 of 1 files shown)


pycaqtimage/pycaqtimage.sip
ZLLentz, line 1: I realized last night that this fix is incomplete.

This does let us iterate through the epics array with the correct sizing information.
However, the result is still packed into a QImage with only one byte available for color, so if the color values of our array exceed 255 then there will be aberrations in the image.

pycaqtimage/pycaqtimage.sip
```diff
@@ -460,6 +453,31 @@ static void _pyColorImagePvCallback(void* cadata, long count, size_t size, void*
     }
 }
 
+static void _pyColorImagePvCallback(void* cadata, long count, size_t size, void* usr)
+{
+    ImageBuffer*  imageBuffer = reinterpret_cast<ImageBuffer*>(usr);
+
+    if (count != imageBuffer->size * 3) {
+        // Wrong data size, unsafe to continue
```
tangkong: Do we want to print/report something like we do in the fallback case below?

pycaqtimage/pycaqtimage.sip
tangkong: I wonder what the actual range of bit-depths are out in the field.  I'm inclined to let it fly until we run into an edge case, but I'm also not sure how one might detect the size of the color bit.

tangkong: Is this not something the IOC reports?  Or do we have to build our own logic for inferring this

ZLLentz: The IOC reports its bit depth but it does this in an inconsistent way for color cameras.
I think what I want to do here is:

- Remove the case for 32-bit colors (size=4) because this might crash the RGB accumulator
- Tell people to set their color output modes to 8-bit (which is what a typical monitor supports anyway)

ZLLentz: Note that even with an 8-bit color output you might still have size 2 arrays (16 bit) because the array is pre-allocated expecting the worst-case biggest data type that might be given

pycaqtimage/pycaqtimage.sip
ZLLentz: I think you're right, this is done in a few places in the file. I think something is done about this in the python layer too, I'll double-check and make sure we're being consistent.

ZLLentz: Here's where this is reported to the user

https://github.com/pcdshub/camviewer/blob/d57c135631aac0c7f1168bff3d84a3807486e4a2/camviewer_ui_impl.py#L2076-L2087

I'll still add something here (and in the other check) to make sure info about this ends up in the logs

ZLLentz:
For reference: here is what 16 bit color looks like (I will request people use 8 bit color for now)

Dark areas look grey, light areas look rainbowy

image

slactjohnson (approved):
This makes sense to me, thanks for improving the notes and error `printfs`.

tangkong (approved):
LGTM

PR #39, opened 2025-10-07
MNT: include c-reviewers for the sip plugin and makefile

ZLLentz:
So it's not just Robert and I getting pinged when something gets updated here

I decided to use the underutilized c-reviewers group
I thought the Makefile needed attention from both c-reviewers and from python-reviewers because it relates to both frameworks


tangkong (approved):
LGTM

PR #40, opened 2025-10-15
FIX: disable all sub-scrolling behavior inside the right hand scroll area

ZLLentz:
It's common for scientists to scroll down the right hand size of the GUI which currently can change settings, orientations, etc. and generally be a menace.

This applies a no-scrolling-allowed filter to every widget inside the right-hand scroll area widget. The scroll widget can still be scrolled with the scroll wheel, but the contents will simply be put in focus like any other mouse event instead of moved/changed via scroll.

I applied this broadly to QWidgets instead of selecting specific types or names of widgets so a future maintainer does not need to specify new widget types when more are added.

Ref

- https://jira.slac.stanford.edu/browse/ECS-7935
- https://jira.slac.stanford.edu/browse/ECS-9081

### DIFF
--- a/camviewer_ui_impl.py
+++ b/camviewer_ui_impl.py
@@ -216,7 +216,6 @@ class NoScrollWheelFilter(QObject):
         return False
 
 
-
 SINGLE_FRAME = 0
 LOCAL_AVERAGE = 2
 
--- a/camviewer_ui_impl.py
+++ b/camviewer_ui_impl.py
@@ -48,6 +48,7 @@ from PyQt5.QtWidgets import (
     QMessageBox,
     QSizePolicy,
     QSpacerItem,
+    QWidget,
 )
 
 import param
@@ -201,6 +202,19 @@ class ScrollSizeFilter(QObject):
         elif event.type() == QEvent.Hide:
             self.set_no_scroll_size()
         return False
+
+
+class NoScrollWheelFilter(QObject):
+    """
+    A filter that prevents the scroll wheel from being used to interact with widgets.
+    """
+
+    def eventFilter(self, _, event) -> bool:
+        if event.type() == QEvent.Wheel:
+            # True stops all other event processing
+            return True
+        return False
+
 
 
 SINGLE_FRAME = 0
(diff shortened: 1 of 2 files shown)


camviewer_ui_impl.py
```diff
@@ -716,6 +729,9 @@ def __init__(
         self.ui.rightPanelWidget.setFixedWidth(width)
 
         self.efilter = FilterObject(self.app, self)
+        self.no_scroll_filter = NoScrollWheelFilter(parent=self)
+        for child_widget in self.ui.rightPanelWidget.findChildren(QWidget):
+            child_widget.installEventFilter(self.no_scroll_filter)
```
tangkong: Should we be doing this every time we load a camera?  When you change cameras, sometimes widgets are added/removed, meaning we could potentially miss a widget?

tangkong: Or are they all loaded at  init and we hide the irrelevant ones on camera load?

ZLLentz: Oh no I should check on this

ZLLentz: - Most of the widgets are loaded at init and are shown/hidden
- Two sections are handled differently as per interactive testing and I need to look more closely

ZLLentz: Okay, I think it's correct now- the widgets that move in and out of the scroll area depending on settings are covered, and the widgets that are created/destroyed for the camera controls are covered.

In some cases these widgets may get the filter double-applied, for example if they are present in the scroll area on first load, but I think this isn't a big deal since only the first filter will be run.

camviewer_ui_impl.py
```diff
@@ -716,6 +735,10 @@ def __init__(
         self.ui.rightPanelWidget.setFixedWidth(width)
 
         self.efilter = FilterObject(self.app, self)
+        self.no_scroll_filter = NoScrollWheelFilter(parent=self)
+        self.no_scroll_filter.install_on_children(self.ui.rightPanelWidget)
+        # specificdialog is sometimes in the scroll area, sometimes not
```
tangkong, line 740: ... that's fun to know

tangkong (approved):
LGTM

I did notice that the lens sider bar is still interactable even on psbuild.  I get a write access complaint exactly once, after which the terminal output is silent.  Might be a good future effort to disable that alongside the start/stop buttons

ZLLentz:
> lens sider bar

This is probably buggy in other ways too because of how rarely we configure a camera with a lens slider

camviewer_ui_impl.py
ZLLentz: Yeah, there's an option for putting it into a floating configuration window or for putting it into the right hand area

